Issue #30, opened 2015-08-15
Turn off logging by default


daveyarwood:
Timbre logs a bunch of stuff at the DEBUG level, which is all visible out of the box when you clone this repo and start to play around with Alda. I put a little blurb in the README about how to turn this off, but it would be nice if it were off by default. I'm sure there is a way to do this with Timbre, just need to look for it.

crisptrutski:
Boot-style options of `-v`, `-vv` or `-vvv` for escalating levels of logging could be a nice interface for cranking things back up

daveyarwood:
Brilliant! I might hack that in today. :saxophone: :notes:

PR #52, opened 2015-09-07
Make 'warn' the default log level in production

daveyarwood:
This PR turns off debug-level logging when using Alda either from `alda.cli` (i.e. the `bin/alda` script) or `alda.now`. The default log level is now `warn`, so users will still see warnings and errors.

The default log level for running tests via `boot test` is still `debug`.

In all scenarios, the log level can be overridden by setting the `TIMBRE_LEVEL` environment variable.


daveyarwood:
+1 good job

daveyarwood:
Fixes #30.

PR #55, opened 2015-09-08
[repl] :load, :map and :score commands

daveyarwood:
3 new REPL commands:

`:load <filename>` loads an Alda score from a file. Prompts for confirmation that you want to abandon your current score if the current score is "dirty" (this needs to be improved -- right now it assumes the score is "dirty" if it's not empty)

`:map` pretty-prints the current data state of the score, i.e. the result of evaluating `(alda.lisp/score-map)`

`:score` prints out the current score-in-progress, as Alda code.

Incidentally, refactored a little bit in the alda.repl.commands namespace, and included the REPL ConsoleReader as an argument so that REPL commands can do things like `.readLine` on it.

### DIFF
--- a/src/alda/repl.clj
+++ b/src/alda/repl.clj
@@ -108,7 +108,7 @@
 
             (re-find #"^:" alda-code)
             (let [[_ cmd rest-of-line] (re-matches #":(\S+)\s*(.*)" alda-code)]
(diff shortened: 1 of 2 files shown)


daveyarwood:
:+1: :fireworks: :confetti_ball: :crocodile:

PR #64, opened 2015-09-09
define octave shift direction in grammar

g200kg:


crisptrutski:
Whoops, looks like we plucked the same fruit - #65 

Apologies :sweat_smile:

daveyarwood:
Ha! And at exactly the same time! 

@g200kg I really appreciate your contribution, but will have to go with #65 instead as it updates alda.parser as well. Thanks!

g200kg:
okay, thank you :+1:

Issue #59, opened 2015-09-08
octave-change direction by "<" / ">" should be explicitly defined in BNF


g200kg:
This confusion often occurs in MML-based language

daveyarwood:
I like the idea of explicitly defining `octave-up` and `octave-down` tokens in the grammar. a Pull Request on this would be greatly appreciated (you would become Alda contributor #4!)

daveyarwood:
[alda.parser](https://github.com/alda-lang/alda/blob/master/src/alda/parser.clj) should also be updated to generate the appropriate code for those tree nodes -- it would be more elegant than the current code, which explicitly checks whether the body of the node is ">" or "<"

crisptrutski:
Whoops, should have used `closes` not `refs`

PR #65, opened 2015-09-09
Place octave-(up/down) directly into grammar

crisptrutski:
Noticed something simple, and decided to dip my toes in :swimmer: :smile: 

Refs #59


crisptrutski:
Niggly point on your naming scheme for test namespaces: can't use CIDER commands because it expects `-test` suffix in the namespace name (otherwise assumes you're calling from namespace-under-test, and fails trying to reload the namespace with suffix added)

Being finite in my bravery, I haven't checked whether I can change/configure things on the CIDER side. Would you object to a PR renaming the namespaces and updating the test task, for my creature comfort?

daveyarwood:
Looks good - awesome first PR, thanks!

I'm going to make a small change before merging -- using the `#(list ...` format in alda.parser instead of `(constantly (list ...` for the sake of consistency.

I don't have a particular reason for the current test namespace scheme -- would happily accept a PR rearranging them for the benefit of CIDER users.

crisptrutski:
One different with the `constantly` version is that it's a bit more memory friendly (shares that immutable data), and to split hairs you do use that form a bit above (although for a different group) :smile: 

Thanks for the flexibility, I just might make that change the next time I update any tests

daveyarwood:
Aha - I didn't know that. Maybe we should change them all to `(constantly '(alda.lisp/...` for a minor perf boost.

crisptrutski:
Can only use it for the cases that don't take any parameters - but I'm cool either way. Reached for `constantly` before I figured out that `<...>` could stop the arguments being passed (creating anonymous functions that ignore their inputs isn't that pretty)

PR #67, opened 2015-09-09
Add :play command to play currently loaded score

crisptrutski:
Just hacked this in, after realising it wasn't there and really wanting it :tropical_drink:

### DIFF
--- a/src/alda/repl.clj
+++ b/src/alda/repl.clj
@@ -1,22 +1,17 @@
 (ns alda.repl
   (:require [alda.version                 :refer (-version-)]
-            [alda.parser                  :refer (parse-input)]
             [alda.lisp                    :refer :all]
-            [alda.sound                   :refer (set-up! tear-down! play!)]
-            [alda.sound.midi              :as    midi]
+            [alda.sound                   :refer (set-up! tear-down!)]
+            [alda.repl.core               :as    repl :refer (*repl-reader* 
+                                                              *parsing-context*)]
             [alda.repl.commands           :refer (repl-command)]
-            [alda.now                     :as    now]
-            [instaparse.core              :as    insta]
             [boot.from.io.aviso.ansi      :refer :all]
             [boot.from.io.aviso.exception :as    pretty]
             [boot.util                    :refer (while-let)]
-            [taoensso.timbre              :as    log]
-            [clojure.java.io              :as    io]
             [clojure.string               :as    str]
-            [clojure.set                  :as    set])
+            [taoensso.timbre              :as    log])
   (:import  [jline.console ConsoleReader]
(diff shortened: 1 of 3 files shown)


crisptrutski:
Some things I noticed, possible bugs or behaviour you'd want to redefine:

1. Composing further from REPL will flush the loaded score, rather than append to it
2. :play doesn't discriminate between a score composed at REPL or loaded from file
3. REPL takes commands while playback is occurring, and and plays simultaneously

I'd say that 1 should just append (can always type :new) and 3 should block on playback. 

Having a way to interrupt playback manually would be useful too

daveyarwood:
> 1. Composing further from REPL will flush the loaded score, rather than append to it

I'd be happy to rethink this, but for now what I'm thinking is: composing further from the REPL will always append lines to the current score-in-memory (`*score-text*`). Eventually users will be able to type `:edit` and `*score-text*` (or better yet, the actual score file they are editing) will open in their `$EDITOR` of choice, so they can delete and rearrange lines as they see fit. Then the resulting file would be re-loaded into `*score-text*`. My idea might be a bit clunky, not sure if there might be a better way to handle this.

> 2. :play doesn't discriminate between a score composed at REPL or loaded from file

What if we allowed passing in extra arguments to the `:play` command? Like `:play file /tmp/foo.alda`, `:play score` for the current score? Perhaps the `:play file` variation would not load that score into memory or clobber the score you're currently editing.

> 3. REPL takes commands while playback is occurring, and and plays simultaneously

I originally had it set up so that playback was always synchronous, but I found it annoying having to wait for playback to finish before I could start typing the next command. 

> Having a way to interrupt playback manually would be useful too

I've had that thought -- `:stop` and `:pause` REPL commands would be ideal.

crisptrutski:
Totally onboard with your suggestions - popping out to $EDITOR would be neat! (reminds me of `pry` from Ruby-land)

Not sure about syntax for play, perhaps just `:play` and `:play /tmp/foo.alda`, or `:playback` and `:play-file`

daveyarwood:
I'm working on refactoring this code a bit -- going pretty well so far. I've cleaned up the REPL command business by pulling out useful information like the parsing context and REPL ConsoleReader into dynamic vars in a new shared namespace, `alda.repl.core`, and I moved some of the internals of `alda.repl` into that namespace as well to keep things clean. From there, I was able to move the `:play` task in with the rest of the commands in `alda.repl.commands`

It's mostly working now, but there is a weird bug where `*score-text*` seems to be reset to `""` every time `:play` is called. Will continue to look at this tomorrow, need sleep now :zzz:

daveyarwood:
OK, I've got it! :metal:

PR #75, opened 2015-09-10
Exit with error code when parser fails

MadcapJake:
Currently when the parser fails, it exits with a success code.  This fixes that using `instaparse`'s `failure?` function.  Exiting with an error code here would allow better support for [atom-build](https://github.com/noseglid/atom-build) and could even move the user's cursor to the error in the file (with the help of a regex).

I don't have much experience with clojure, so let me know if there's a more idiomatic way of doing this.

### DIFF
--- a/src/alda/cli.clj
+++ b/src/alda/cli.clj
@@ -8,7 +8,7 @@
             [alda.version    :refer (-version-)]
             [alda.sound]))
 
-(defn fluid-r3! 
+(defn fluid-r3!
   "Fetches FluidR3 dependency and returns the input stream handle."
   []
   (eval
@@ -26,6 +26,9 @@
   (if-not (or file code)
     (parse "--help")
     (let [alda-lisp-code (parse-input (if code code (slurp file)))]
+      (when (instaparse.core/failure? alda-lisp-code)
+        (pprint alda-lisp-code)
+        (System/exit 1))
       (when lisp
         (pprint alda-lisp-code))
       (when map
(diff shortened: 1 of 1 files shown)


daveyarwood:
Good idea -- Thanks for the PR! 

We should move this logic into `alda.cli` though -- the problem with it being in the `alda.parser/parse-input` method is that this method sometimes needs to be called without exiting, e.g. from the Alda REPL.

If I'm understanding correctly, you want to use the `alda parse` task to parse, or maybe even the `alda play` task to play the score you're editing, right? So, we should make it a condition in these two functions: https://github.com/alda-lang/alda/blob/master/src/alda/cli.clj#L20-L56

The `play` function actually already has the `(if (instaparse.core/failure? parsed) ...` check, so you can just replace `(prn parsed)` with `(do (prn parsed) (System/exit 1))`

You'll have to add it to the `parse` task, where you actually have the opportunity to fix a small bug I just noticed, in the process\*. 

\*Currently, if you try and parse some bogus alda code, and print both the `--lisp` and `--map` versions of the parsed code, it prints the Instaparse failure object twice.

```
$ alda parse --lisp --map --code "APPLESAUCE"
{:index 10,
 :reason
 [{:tag :string, :expecting ":"}
  {:tag :string, :expecting "|"}
  {:tag :string, :expecting "#"}
  {:tag :string, :expecting "(*"}
  {:tag :regexp, :expecting #"\s"}
  {:tag :string, :expecting "\""}
  {:tag :string, :expecting "/"}],
 :line 1,
 :column 11,
 :text "APPLESAUCE"}

{:index 10,
 :reason
 [{:tag :string, :expecting ":"}
  {:tag :string, :expecting "|"}
  {:tag :string, :expecting "#"}
  {:tag :string, :expecting "(*"}
  {:tag :regexp, :expecting #"\s"}
  {:tag :string, :expecting "\""}
  {:tag :string, :expecting "/"}],
 :text "APPLESAUCE",
 :column 11,
 :line 1}
```

Instead, it should catch the failure scenario, print the exception once, and then exit with an error status.

MadcapJake:
Ok, made the adjustments, did a few runs with `boot alda -x` and it seems to be working! Let me know if this is what you meant.

daveyarwood:
Perfect! :boom:  Would you mind squashing this down to one commit? Then I think it'll be good to go.

MadcapJake:
done and done! :+1:

PR #80, opened 2015-09-10
Remove 100% constraint on quantization

### LINKED ISSUE #76: Overlapping notes with same voice
Perhaps there's already a convenience way to note this, or it's the job of some post-processing, but I'd like to be able have overlap between notes, eg. play a series of 8th notes all with a 4th note duration.

Examples would be for guitar or harp lines.

As far as I can tell, the only way to do this would be to have multiple voices and alternate between them, but it would be nice to just set the duration separately and write the notes serially.

crisptrutski:
Closes #76

### DIFF
--- a/src/alda/lisp/attributes.clj
+++ b/src/alda/lisp/attributes.clj
@@ -11,6 +11,10 @@
   {:pre [(<= 0 x 100)]}
   (constantly (/ x 100.0)))
 
+(defn- unbound-percentage [x]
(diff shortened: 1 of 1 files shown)


crisptrutski:
Like this suggestion, works a treat :guitar:

daveyarwood:
:books: :tractor: :round_pushpin: :kr: :mortar_board:

PR #81, opened 2015-09-10
Test reorganisation

crisptrutski:
Decided to just knock this off while adding a test for the quantization


crisptrutski:
@daveyarwood `alda.test-helpers` is certainly much clearer, this should be g2g

daveyarwood:
Cool - merging it!

Issue #45, opened 2015-09-07
Support quarter tones / microtonal music


infojunkie:
Would be great to support quarter tones, or microtones in general!

daveyarwood:
This is very high on my priority list.

Some great ideas were proposed [in this Reddit thread](https://www.reddit.com/r/programming/comments/3jvfy4/alda_a_music_programming_language/cusntqi)

Omega9:
Do you plan to support .scl?

daveyarwood:
@Omega9 That's not a bad idea! 

An open question I have for supporting microtonal music in general is how we will express each note, for scales with more than 7 notes (`a`-`g`). It would be ideal if we could continue to express a note as a single character. Currently the Alda parser reserves the letters `a`-`g` for notes, and the letter `r` for rests. 

Some ideas:

- We could enforce an upper limit of 17 notes for scales, so we could use the letters `a`-`q` (since `r` is off limits, and skipping over `r` might be confusing)
- We could use uppercase letters in some way. Or maybe some combination of upper- and lower-case letters.

crisptrutski:
So I'm wanting to take a stab at adding [tonality diamond](https://en.wikipedia.org/wiki/Tonality_diamond) style tunings, and that's informing my thoughts around the literals for notes.. A part of me thinks that tuning "plugins" should be able to bring in their own, intuitive notation.

eg. rather than mapping `a`-`q` to a rhombus, or using a complicated scheme of accidentals-that-interact-with-tuning-model-in-a-fancy way.. you might want to use a 2d dimension system.. a bit like an index into a hexagon grid.

Definitely going to play with Scala this weekend (this is worse than Clojure / Closure..)

daveyarwood:
@crisptrutski Whoa... this tonality diamond stuff looks crazy! Hardcore experimentalists will love this kind of thing, if we bring it into Alda. I would experiment with it, too.

I like the idea of leveraging the plugin system for tunings, rather than hard-coding them into Alda as attributes. 

I'm not sure yet exactly how the plugin system will work, but I imagine it will involve transforming the score into syntax that the core Alda parser knows how to read -- if that's the case, then we could have tuning plugins that generate Alda code where each note has a precise pitch in Hz -- this is a more general feature I had in mind, where you could override the current octave/note system and just say, e.g. "play a note at 3000 Hz."

crisptrutski:
[...]

PR #84, opened 2015-09-10
Extensible tuning system

crisptrutski:
Played around a bit with tuning, related to #45 

Certainly wouldn't expect this merged as is, don't think the syntax is correct at all.

1. Need more power and extensibility for setting tuning than just a name - eg. tempering around another key, or brute force setting each note

2. Being able to provide wholesale new tuning systems in plugins, eg. that have irregular, asymmetric or non-abelian treatment of accidentals, or octaves that don't double frequency (like a skewed mean-temperament that preserves the 3:2 5th)

3. Introducing tuning parameters, and making the tuning map extensible are both simple enough (pass whole context to tuning, and maybe have a `defattribute` style macro for tunings) - but what seems slightly harder is designing the BNF grammar to be flexible enough, but still helpful.

  Wondering about the maintenance and DRY side of having all these different attributes and their types in the grammar. Thinking that a free-for-all `(<attribute-name> [^)]*)` grammar plus dynamic validation during parsing could be a better fit - data passed to `defattribute` macro, and potentially `deftuning` macro too, could be used to still validate, but suddenly the system is far more open.

4. Lastly, I couldn't find any suitable Beethoven (7th would be great) in alda format to test this tuning out with - should we raise another issue for that? :violin: 

5. Have no idea where to start notating Harry Parch or Terry Riley into alda (or how they even notate things traditionally). This sounds like lots of fun, and I'm keen to try or die learning

### DIFF
--- a/grammar/alda.bnf
+++ b/grammar/alda.bnf
@@ -22,6 +22,7 @@ chord                   = (note | rest) subchord+
 note                    = pitch duration? <ows> slur?
 rest                    = <"r"> duration? <ows>
 
+str                     = #"\w+"
 pitch                   = !name #"[a-g][+-]*"
 duration                = note-length <ows> (tie <ows> note-length <ows>)* slur?
 note-length             = number dots?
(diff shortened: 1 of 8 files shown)


crisptrutski:
Certainly don't have the ear for this stuff, but thought that tuning sounded even *darker* than it should. Started playing with pyhagorean tuning next, and it got worse - unisons.

So midi notes can't be fractions I guess :laughing: 

Searching around a little bit now.. seeing instructions for specific synths to retune them globally using MIDI.. really? :disappointed:

crisptrutski:
Going to try use the pitch wheel..

crisptrutski:
Will come back to this and try apply http://www.midi.org/techspecs/midituning.php rather, this pitch bending is weird :dancers:

daveyarwood:
This may come in handy: http://stackoverflow.com/a/25548933/2338327

Apparently being able to tune individual notes is only a recent development, added sometime around JDK 6 or 7. Before that, you had to use pitch bend, which bends the whole channel, and apparently is a little unwieldy and doesn't work in realtime. Fingers crossed that MIDI Tuning gives us what we want!

crisptrutski:
Thanks, I think this is the same approach that I linked to above - main issue is the limit on the number of notes. Super fancy tuned pieces may need to dynamically reset the tuning, and the complexity and time sensitivity there scares me.

Maybe the best is only to even attempt supporting super-perverse tunings with synths (and wait for MIDI to be replaced :stuck_out_tongue:)

daveyarwood:
I'm going to close this for now -- this branch is too far out of sync with master at this point. A number of things have changed, especially with the major rewrite I have coming very soon™ (take a peek at the `housekeeping` branch if you're curious).

There is lots of great progress here, though -- thanks @crisptrutski for kicking it off! This PR will be a great reference whenever development picks up on issues #45 / #87.

PR #88, opened 2015-09-12
Support "from" and "to" for playing

crisptrutski:
Not quite done with this yet, but perhaps ready for feedback.

There are some performance-oriented things here, like making some functions lazy, a lower-overhead trick than `(doall (pmap ..))` and using transducers in some places.

Will be back this evening to finish this off - supporting labels, typing in with CLI and REPL commands, and adding tests

### DIFF
--- a/src/alda/sound.clj
+++ b/src/alda/sound.clj
@@ -27,8 +27,8 @@
    e.g. for MIDI, create and open a MIDI synth."
   [audio-type & [score]]
   (if (coll? audio-type)
-    (pdoseq [a-t audio-type]
-      (set-up! a-t score))
+    (doall
+      (pmap #(set-up! % score) audio-type))
     (when-not (set-up? audio-type)
       (set-up-audio-type! audio-type score)
       (alter-var-root #'*active-audio-types* conj audio-type))))
@@ -54,8 +54,8 @@
    added to the score between calls to `play!`, when using Alda live.)"
   [audio-type & [score]]
   (if (coll? audio-type)
-    (pdoseq [a-t audio-type]
-      (refresh! a-t score))
+    (doall
+      (pmap #(refresh! % score) audio-type))
     (when (set-up? audio-type)
       (refresh-audio-type! audio-type score))))
 
@@ -77,8 +77,8 @@
    e.g. for MIDI, close the MIDI synth."
   [audio-type & [score]]
   (if (coll? audio-type)
-    (pdoseq [a-t audio-type]
-      (tear-down! a-t score))
+    (doall
+      (pmap #(tear-down! % score) audio-type))
     (when (set-up? audio-type)
       (tear-down-audio-type! audio-type score)
       (alter-var-root #'*active-audio-types* disj audio-type))))
@@ -174,12 +174,14 @@
         events      (shift-events events start end)
         duration    (- (or end (score-length score))
                        (or start 0))]
-    (pdoseq [{:keys [offset instrument] :as event} events
-            :let [instrument (-> instrument instruments)]]
-      (at (+ begin offset)
-          #(when @playing?
-             (play-event! event instrument))
-          pool))
+    (doall
+      (pmap (fn [{:keys [offset instrument] :as event}]
+              (let [instrument (-> instrument instruments)]
+                (at (+ begin offset)
+                    #(when @playing?
+                       (play-event! event instrument))
+                    pool)))
+            events))
 
     (when-not async?
       ; block until the score is done playing
--- a/src/alda/sound.clj
+++ b/src/alda/sound.clj
@@ -2,7 +2,7 @@
   (:require [alda.sound.midi :as    midi]
             [overtone.at-at  :refer (mk-pool now at)]
             [taoensso.timbre :as    log]
-            [alda.util       :refer (check-for)]))
+            [alda.util       :refer [check-for parse-time pdoseq parse-position]]))
 
 (def ^:dynamic *active-audio-types* #{})
 
@@ -15,7 +15,7 @@
 
 (defmethod set-up-audio-type! :default
   [audio-type & [score]]
-  (log/errorf "No implementation of set-up-audio-type! defined for type %s" 
+  (log/errorf "No implementation of set-up-audio-type! defined for type %s"
               audio-type))
 
 (defmethod set-up-audio-type! :midi
@@ -27,9 +27,9 @@
    e.g. for MIDI, create and open a MIDI synth."
   [audio-type & [score]]
   (if (coll? audio-type)
-    (doall 
-      (pmap #(set-up! % score) audio-type))
-    (when-not (set-up? audio-type) 
+    (pdoseq [a-t audio-type]
+      (set-up! a-t score))
+    (when-not (set-up? audio-type)
       (set-up-audio-type! audio-type score)
       (alter-var-root #'*active-audio-types* conj audio-type))))
 
@@ -40,7 +40,7 @@
 
 (defmethod refresh-audio-type! :default
   [audio-type & [score]]
-  (log/errorf "No implementation of refresh-audio-type! defined for type %s" 
+  (log/errorf "No implementation of refresh-audio-type! defined for type %s"
               audio-type))
 
 (defmethod refresh-audio-type! :midi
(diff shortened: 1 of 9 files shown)


daveyarwood:
Awesome! I will take a closer look tomorrow.

crisptrutski:
@daveyarwood this is basically working now - left my blindly-stumbling trail in case it helps the changes make more sense, but let's squash before merging.

I think we should choose between `from/to` and `start/finish` - got one pair from the existing code, and one pair from the issues and they met up, all mismatched, in the middle.

The `set/difference` bothers me a bit for the repl - feel like we should be able to skip building expensive sets, doing diffs, and transforming all the maps by just doing offset math direct into the `play!` inner loop, and keeping track of "last-played-offset" in REPL state rather - but that's slightly out of scope here.

An example of `set/difference` being weird - defining an event in the past, by using `at-marker`

daveyarwood:
OK -- today is my day to nurse my Hopscotch hangover and catch up on Alda issues/PRs. Will be taking a closer look at this one later today. 

One quick comment: I agree about choosing between from/to and start/finish so that it's unified. I like from/to better -- it reads more like prose.

daveyarwood:
Re: the Clojure 1.7.0 features you're using here, I would say let's leave them in, and we'll just make sure that we fix the Boot/Maven dependency issue before we merge this -- it's probably less work that way, and besides, I've reached the point where I'm not satisfied with settling for 1.6.0 compatibility anymore :triumph:

daveyarwood:
In the process of reviewing this PR now -- so far so good!

> The set/difference bothers me a bit for the repl - feel like we should be able to skip building expensive sets, doing diffs, and transforming all the maps by just doing offset math direct into the play! inner loop, and keeping track of "last-played-offset" in REPL state rather - but that's slightly out of scope here.

> An example of set/difference being weird - defining an event in the past, by using at-marker

I knew my use of `set/difference` felt a little weird when I wrote it. I would definitely be open to rewriting it in a way that makes more sense, both performance-wise and guarding against the possibility of getting into a weird state with markers, etc.

src/alda/sound.clj
```diff
@@ -121,38 +134,66 @@
 
 (def ^:dynamic *play-opts* {})
 
-; TODO: control where to start and stop playing using the start & end keys
+(defmacro with-play-opts
+  "Update *play-opts* w"
```
daveyarwood: It looks like you may have been kidnapped by [Candlejack](http://knowyourmeme.com/memes/candlejack) while you were wri

crisptrutski: [...](http://i1.kym-cdn.com/photos/images/original/000/045/974/437px-Clarinet_boy_seen_things.jpg)

crisptrutski:
Hope the scotch & hops hangover has passed! I'll fix up that docstring and consolidate the naming tomorrow, refining the event-diff in a later pull request. Any other crits to tackle?

daveyarwood:
@crisptrutski That sounds good to me! 

The only other thing in the way is the Clojure 1.7 features... I'm worried that it will cause Alda not to work for people using a version of Boot older than 2.2.0, or with `BOOT_CLOJURE_VERSION` set to something < 1.7.0. I'm still on the fence about whether this is something we need to fix now (i.e. #92 ), or if we can just strongly advise that people use Boot 2.2.0+ and not have `BOOT_CLOJURE_VERSION` set to < 1.7.0

pkobrien:
I don't see a problem with requiring Boot 2.2.0 or greater, and Clojure 1.7 or greater.

crisptrutski:
@daveyarwood beep beep :rotating_light:!

Placed a marshmallow on top of this code, amended as we discussed, reheated it a bit, and squashed it down to edible proportions. Hopefully you find it to your taste :smile: 

Late thought bubble: it would be nice to have some more "end-to-end"-ish tests over this functionality. Maybe `(is (= expected-event-map (eval (parse-with-play-opts alda-code some-suitable-opts))))`? Perhaps a whole slew of such tests would be useful.

daveyarwood:
Suhweeet! This looks really good. I'll play around with it later tonight and hopefully merge it. I think I'm just gonna take the plunge and start including Clojure 1.7 features. I don't think it's too much to ask for people to use >= Boot 2.2.0, and eventually Alda should be self-contained anyway, so that users won't even need to have Boot.

Hmm... re: the end-to-end tests, do you think it might be too much of a hassle to keep updating the expected event map each time we add a new feature? The instruments, note events, etc. are going to steadily accrue more and more fields over time.

crisptrutski:
While few things are more painful things dragging overly particular tests up the hill to your latest commit..

> Wiping away how your version of clojure chooses to order the keys in that map literal...
> Patiently informing every single AST node that it's had a new daughter..
> Ensuring it won't run less than `duration-last-time * fudge` before next leap second...

... I'm still generally a fan of end-to-end tests.

Perhaps it's a bit early to have any, or perhaps they're not a good idea for alda at all - but clobbering expectations with new output (or deleting that test) get you out of trouble pretty quickly!

Let's hang ten if you're not feeling the pull - time spent not doing what you think you'll need comes in really handy when you actually need things :smile:

daveyarwood:
I think a better idea might be to write particular tests that don't check for absolute equality, but rather for very specific things in the resulting score-map.

e.g. 

    (is (= (set (map :offset (:events score-map)))
           #{500.0 1000.0 1250.0}))

That being said, though, if you were to feel like slipping some more end-to-endy tests in with a PR, I wouldn't object :)

daveyarwood:
I found a few issues, testing this with `test/examples/awobmolg.alda`:

- I tried running `boot alda -x 'play --file test/examples/awobmolg.alda --from 0:02'` and the same thing with `--from 0:30` -- for both, there was a long pause, and then it returned without playing anything.

- I can't seem to use markers from the `alda play` task -- I got this exception when I tried both `boot alda -x 'play --file test/examples/awobmolg.alda --from voiceIn'` and `boot alda -x 'play --file test/examples/awobmolg.alda --from :voiceIn'`:

        clojure.lang.ExceptionInfo: java.lang.NumberFormatException: For input string: "voiceIn"
        ...
        java.util.concurrent.ExecutionException: java.lang.NumberFormatException: For input string: "voiceIn"
        java.lang.NumberFormatException: For input string: "voiceIn"
         sun.misc.FloatingDecimal.readJavaFormatString              FloatingDecimal.java: 2043
         sun.misc.FloatingDecimal.parseDouble              FloatingDecimal.java:  110
                                          ...
                      alda.util/parse-time/fn                          util.clj:   30
                          clojure.core/map/fn                          core.clj: 2624
                                          ...
                             clojure.core/seq                          core.clj:  137
                       clojure.core/concat/fn                          core.clj:  691
                                          ...
                         alda.util/parse-time                          util.clj:   28
                       alda.sound/lookup-time                         sound.clj:  151
        ...

- After loading the score from the Alda REPL, `:play from :voiceIn` works but `:play from voiceIn` throws the same exception as above -- I think we should support doing both ways, and maybe also `:play from @voiceIn`

- Unrelated, but I just noticed there is a lingering debug prn statement [here](https://github.com/alda-lang/alda/blob/master/src/alda/repl/commands.clj#L112) when using the `:load` task, probably my fault -- I'm going to fix that on master now :P

crisptrutski:
Yipes! Did some cursory REPL playback tests after find-replacing last night, but perhaps there were some stale definitions, or I just missed something.

One thing I've noticed due to the in-ns trick is that cider gets confused and opens a temp file instead of the real file when using goto-definition. Perhaps I saved some necessary changes into the aether...

Clojure itself uses a different trick for breaking up namespaces, using load-file I think, but don't swe any point addressing this until we find the CLJC suitable solution.

Have quite a busy week remaining, will likely only be able to fix this over the weekend, but feel free to take the wheel if you want to roll this in earlier :truck:

crisptrutski:
Oh yes, I only supported the keyword-string style. Let's rather check against a looks-like-timestamp regex.

crisptrutski:
OK, updated to handle de`colon`ised markers, and fixed some other small things. The "loads for some time, then plays nothing" sounds like.. clock offset issues (loading time is being counted?), will look at that now

crisptrutski:
@daveyarwood Sorry I didn't have time to finish this over the weekend, only problem left is the.. um.. hearing-absolutely-nothing issue.

I think my `pdoseq` was a bit too clever for it's own good, and initialisation of timing of `refresh!` and playback are a bit goofed. 

If you're impatient, could revert the `pmap` stuff and I believe things should work OK

daveyarwood:
@crisptrutski No rush! I'll leave it up to you whether you want to revert to `pmap` or leave `pdoseq` in. I kinda like `pdoseq`, for what it's worth.

Have you tried turning on debugging with `TIMBRE_LEVEL=debug` and seeing if the notes are even being generated? That might help isolate where the problem is.

daveyarwood:
I have a hunch that the notes not playing may be related to the (unrelated) bug I fixed incidentally in #102 ... I'll do some testing.

crisptrutski:
@daveyarwood reverted `pdoseq` usage for now

crisptrutski:
Figured it out - the blocking variant was not respecting the `:let` clause. 

Still has sharp edges - if iterating through more than one collection the latch will release too early. Perhaps just asserting that it's not used in this way is enough.

Anyhoo - should be mergable now :smile:

daveyarwood:
Awesome, works like a charm! Totally mergeworthy :+1:

Issue #36, opened 2015-09-06
Feature request: Set the key


FragLegs:
It would be nice to be able to set the key (either by name or as a series of flatted/sharped notes) as an attribute and global attribute. I don't know Clojure well enough to quickly set up a pull request for this, but if I find time I might try my hand at it. In the meantime, I put this out there for anyone else working on it.

daveyarwood:
This would be a step in the right direction toward eventually being able to export sheet music for an Alda score. 

I think what we want is to define a new top-level thing called `*key*` that is initialized as an empty map `{}` when a score is initialized, and any time the key is set via an attribute/global attribute change, the current time marking and the key is recorded in `*key*`. What would be even more useful is if we could store the current measure number, but we'd have to implement measures first. (currently, Alda just ignores bar lines)

I don't think we'll be able to do much with the key-change data for the time being, but once things get moving with the sheet music export feature, we'll be able to use the key-change data to tell it what key signature to display.

crisptrutski:
Perhaps this would even change the default meaning of bare notes like `c` - with a "natural" modifier to escape the key like `c.`? This would dovetail into microtonal or custom tunings - perhaps "keys" could even introduce new note symbols within their scope - like `aa` or `h` or some suitable unicode :notes:

FragLegs:
Yeah, that's more along the lines of what I was thinking. I just don't want
to have to type "b-" for every B flat in a song when the song is in the key
of B flat. Motivated by the same laziness that invented key signatures in
the first place. :)

daveyarwood:
Aha! I see what you're saying now. I think that's a great idea! Disregard my last comment -- on second thought, I think we should keep track of keys on a per-instrument basis, like everything else. Being able to have 2 instruments playing at the same time in 2 different keys is totally in line with Alda's way of doing things. For example, you can have different tempos for different instruments by doing something like this:

    violin: (tempo 180) e f g e f g e f g e f g e f g
    viola: (tempo 200) e f g e f g e f g e f g e f g

which creates a nice phase effect. To change the tempo for all instruments, you can do it globally with the `tempo!` global attribute.

We could do a similar thing with keys -- allowing you to change the key per-instrument with `key` or for every instrument with `key!`. An instrument's key can be C major by default, so if you don't provide an accidental for a note, it will be natural by default, which is the way it works currently.

[...]

PR #89, opened 2015-09-12
Key signatures

FragLegs:
My initial attempt at providing the feature referenced in issue #36. I need to fix all of the now-broken unit tests, add new ones, and provide documentation, but I wanted to open the PR early to get feedback on the design.

You can change a key signature like this `piano: (key c++ e-) c c+ c=` which will map to `c++ c+ c`

This is my first time working in Clojure, so please tell me what you don't like and why - it will help me immensely in my learning. I'm sure I did a lot of stupid noobish things.


crisptrutski:
This looks pretty good to me - the approach looks sounds and I like the parser refactoring. 

No big "what the :hankey:" moments here for me (but perhaps we'll both get schooled by a bigger fish) :smile:

Here's feedback, mostly style fluff (since you asked!) - only the first point would affect end users, so feel free to skim if it's overly pedantic :zzz: 

1. Perhaps we want some validation on the pitch forms (natural never mixed with sharps/flats) and key forms (no letter is present more than once).

  If we decided to allow mixing naturals, I'd expect them to work slightly differently - with `=` being a reset rather than identity (ie `c++=` is `c` and `c+=+` is `c+`).

2. In addition to fixing and adding appropriate tests - it would be nice to add a piece to examples that uses this. It's good to show off features and make them more discoverable, they're like integration tests too. Would suggest something which switches key at least once also.

2. Watch out for whitespace - trailing spaces, hanging braces, indenting etc. Pretty standard advice lives [here](https://github.com/bbatsov/clojure-style-guide), but just configuring the appropriate plugin for your editor is probably the easiest

3. Not sure if you meant to leave the `log/debug` statement. I've been avoiding adding any myself, maybe some guidance from @daveyarwood would help contributors know when it's appropriate.

4. Nice change from `case` to `map as the fn` - just note that this can allow `nil`s to seem through (grammar has us covered here, but just mentioning the trade-off in case you hadn't thought of it. `nil` chasing can be frustrating in large code bases)

5. `check-key` function is poorly named in my opinion, perhaps `apply-key` shows purpose (it's a transformation not a predicate) better?

6. `(fn [val] (constantly (parse-key-signature (rest val)))))` could be `(comp constantly parse-key-signature rest)`. Sometimes chasing "one-liners" hurts readability, but in this case I think it's clearer as well as shorter. On a side note, it's so common to add `constantly` with `defattributes`, I wonder if the DSL shouldn't give some sugar here.

7. The signature change and immediate sort-of-destructing in `pitch`:  [parsed-pitch] (let [letter (first parsed-pitch) accidentals (rest parsed-pitch)]` seems like a strange change, I'd just change the signature back and skip the `let` altogether. For `let` statements like that it's more idiomatic to use destructuring too - but always hoist up to the enclosing function if it makes sense.

8. `(get signature letter nil)` - more idiomatic to just leave out the `nil` at end (it's implicit)

9. `(into {} (map (fn [pitch] {(first pitch) (rest pitch)}) pitches))` has a few things to speak to:
  
  Rather use `[key value]` than `{key value}` in the mapping function when building a map - it's slightly cheaper and more commonly seen - vectors are the "tuples" of Clojure. It also gives nice symmetry - when iterating over maps the key-value pairs are represented as vectors.

  Destructuring instead of calling both `first` and `rest` would be idiomatic here again. 

  Finally you could use a transducer here instead of actually building the lazy sequence. It doesn't matter here at all, as the data is small, and this runs before playback, so we don't care much about the few extra objects being allocated. Just mentioning as it's such a simple change to just transdue- probably a good habit to get into for this codebase. GC pauses in the middle of co-ordinating sounds can be quite noticeable. 

  (@daveyarwood I saw an old commits for `1.6` support (`update-in` vs. `update`) - are you still wanting to support that, or was that just convenience before `boot` upgraded its default?)

  Putting all three points together, I'd have written it like so: `(into {} (map (fn [x & xs] [x xs])) pitches)`, perhaps even using `letter` instead of `x` for some extra context.

daveyarwood:
Nice, Shayne! Will dig in on this more later today. I have one quick comment, so far:

> check-key function is poorly named in my opinion, perhaps apply-key shows purpose (it's a transformation not a predicate) better?

+1 on this. `apply-key` would be a good name for this function

FragLegs:
Thank you! This is exactly the kind of feedback I was looking for. So many little things to learn in new languages. :)

With regard to validation - I agree that some validation there would be nice, but I also want to avoid being overly prescriptive about what can and can't be done with the language. Would a warning be more appropriate for these things that don't break but aren't necessarily the expected behavior?

daveyarwood:
@crisptrutski 

> (@daveyarwood I saw an old commits for 1.6 support (update-in vs. update) - are you still wanting to support that, or was that just convenience before boot upgraded its default?)

I would like to move away from Clojure 1.6 support, given all the useful things that Clojure 1.7 provides (transducers, etc.). Clojure 1.7.0 is in fact already a dependency in the build.boot, I just reverted the `update` to `update-in` because I got a couple reports of seeing `java.lang.RuntimeException: Unable to resolve symbol: update in this context` when trying to run Alda, presumably with a version of Boot older than 1.6.0, or perhaps with `BOOT_CLOJURE_VERSION` set to 1.6.0 locally. I wasn't sure how best to handle this, so reverting to `update-in` was my way of putting a band-aid on the problem.

I'd like to tackle this for real now and enforce the Clojure 1.7.0 dependency in a way that won't break when running the `alda` executable from outside of the project directory. This makes my head hurt a little bit, so we should discuss this in issue #92 and figure out a solution before we start adding Clojure 1.7.0 syntax to the Alda codebase.

crisptrutski:
Could add a step to `build.boot` that aborts with a friendly message if clojure is out of date - it's quite a stumbling block that setting in `(set-env! ..)` is not always enough (Maven dependency resolution seems so drunk to me - pull in some ranges from a transitive dependency and I'll make the worst choices)

crisptrutski:
Let me know if I should back all the transducer stuff out of this PR

src/alda/lisp/model/pitch.clj
```diff
@@ -17,17 +17,38 @@
...
-        (midi->hz midi-note)))))
+        (midi->hz midi-note))))))
+
+(defn parse-pitch 
+  "Converts the tokenized representation of a pitch into a keyword sequence.
+   For example, ('c' '+' '+') will become (:c :sharp :sharp)"
+  [letter & accidentals]
+  (list* (keyword (str letter))
+        (map {\+ :sharp, \= :natural, \- :flat} accidentals))
+)
```
daveyarwood: This is a nice refactoring - not bad for a budding Clojurist! 

It feels a little off to have a code-generating function in this namespace, though -- it's more of an implementation detail for how the parser transforms Alda code into the data we need, vs. a public function that you would theoretically use if you were to write an Alda score in alda.lisp form. In other words, if you in the alda.lisp namespace in a REPL and you were using the functions available in that namespace to construct a score, it would make more sense to represent an E-double-flat as `(pitch :e :flat :flat)` than as `(pitch (parse-pitch "e" "-" "-"))`

I would move this function into `alda.parser` and then have the transform function for `:pitch` use it internally, so that it's already transformed by the time it's turned into code.

grammar/alda.bnf
```diff
@@ -22,7 +22,8 @@ chord                   = (note | rest) subchord+
 note                    = pitch duration? <ows> slur?
 rest                    = <"r"> duration? <ows>
 
-pitch                   = !name #"[a-g][+-]*"
+pitch                   = pitch-notation
```
daveyarwood: could change this to `<pitch-notation>`, so that the contents of the `pitch-notation` node get lifted up into the `pitch` node

grammar/alda.bnf
```diff
@@ -22,7 +22,8 @@ chord                   = (note | rest) subchord+
...
+pitch-notation          = !name #"[a-g][+-=]*"
```
daveyarwood: could change THIS to `<pitch-notation>`, so that the contents of the `:pitch-notation` node get lifted up into the next higher level of the parse tree, e.g. into the `:pitch` node

src/alda/parser.clj
```diff
@@ -20,16 +20,12 @@
...
           :duration          #(list* 'alda.lisp/duration %&)
-          :pitch             (fn [s]
-                               (list* 'alda.lisp/pitch
-                                      (keyword (str (first s)))
-                                      (map #(case %
-                                              \- :flat
-                                              \+ :sharp)
-                                           (rest s))))
+          :pitch             #(list* 'alda.lisp/pitch %&)
+          :pitch-notation    #(list* 'alda.lisp/parse-pitch %)
```
daveyarwood: Then, (following the change I proposed above in `alda.bnf`), you could get rid of the `:pitch-notation` transformation here (since that node will no longer explicitly exist), and have the `:pitch` transformation be `#(list* 'alda.lisp/pitch (apply parse-pitch %&))`

src/alda/lisp/attributes.clj
```diff
@@ -74,3 +74,15 @@
...
+  "Transforms a key signature ((:c :sharp :sharp) (:d :flat)) into a letter
+   to accidentals map {:c (:sharp :sharp), :d (:flat)}."
+  [pitches]
+  (into {} (map (fn [pitch] {(first pitch) (rest pitch)}) pitches)))
+
+(defattribute key-signature
+   "The key in which the current instrument is playing."
+   :aliases [:key]
+   :initial-val {}
+   :transform (fn [val] (constantly (parse-key-signature (rest val)))))
```
daveyarwood: There are a couple things I don't understand here (possibly could be bugs?):

- In Clojure, [`constantly`](https://clojuredocs.org/clojure.core/constantly) is mostly used in the place of an anonymous function -- it returns an anonymous function that will always return a particular value. Because we want a function that returns a value, not a function that returns a function, I think either this shouldn't be wrapped in an anonymous `fn`, or the `constantly` shouldn't be there. Ideally this line should just be `:transform parse-key-signature`, and `parse-key-signature should be modified so that it ignores the first argument.

- I'm not following what the first (ignored) argument actually is :)  Why is `rest` necessary here?

daveyarwood:
> Perhaps we want some validation on the pitch forms (natural never mixed with sharps/flats) and key forms (no letter is present more than once).

---

> With regard to validation - I agree that some validation there would be nice, but I also want to avoid being overly prescriptive about what can and can't be done with the language. Would a warning be more appropriate for these things that don't break but aren't necessarily the expected behavior?

I think in this case, because it's not really clear what should happen when you mix naturals with sharps/flats (something you would never do in traditional music notation), we should restrict people from doing it. A good idiom for validating arguments in Clojure is to use [preconditions](http://blog.fogus.me/2009/12/21/clojures-pre-and-post) -- you can see examples of this with some of the other attributes in [alda.lisp.attributes](https://github.com/alda-lang/alda/blob/master/src/alda/lisp/attributes.clj#L10-L16)

src/alda/lisp/attributes.clj
FragLegs: I think you're right that it is a bug. The first argument is :key-signature and I have no idea why. My limited Clojure ability can only superficially parse what is happening in the defattribute macro, so I was having trouble tracing what should be happening.

src/alda/parser.clj
FragLegs: The reason that I added :pitch-notation as a first class member of the parse is that I needed it in the key signature as well. If I were to take this approach, wouldn't I also have to parse the :pitch-notation in parse-key-signature()? I could certainly do that, but it feels like a lot of parse action happening to far along in the process. I may be misunderstand the flow though.

src/alda/lisp/model/pitch.clj
FragLegs: That makes sense. I wonder if I should also move the parsing of the key signature out into the parser? That is, would it make more sense in the REPL to write (key-signature '(:e :flat :flat) '(:c :sharp)) or (key-signature {:e '(:flat :flat), :c '(:sharp)} ?

[that may not be how you actually use this in the REPL -- I haven't played around with it that way yet]

src/alda/parser.clj
daveyarwood: The parsing library we're using, Instaparse, allows us to [hide tags](https://github.com/Engelberg/instaparse#hiding-tags), which is what I'm proposing here -- each node that needs to capture a `pitch-notation` node can still do so -- the content of the `pitch-notation` node will just be pulled up into the next highest level.

e.g. "c++" will parse as `(pitch :c :sharp :sharp)` instead of `(pitch (pitch-notation "c" "+" "+")`

We would doing the pitch-notation parsing earlier on, and making it more of a implementation detail as part of the parsing process.

src/alda/lisp/model/pitch.clj
daveyarwood: I think doing more on the parsing side is generally a good thing -- so far I've been trying to leverage Instaparse's powerful [transformations](https://github.com/Engelberg/instaparse#transforming-the-tree) as much as possible, to help keep the rest of the code cleaner and easier to understand. The idea is that the parser picks up on a pattern of text as a particular type of node, then transforms that type of node into the type of data that we actually want to work with.

It gets a little more interesting with Alda, because the type of data that we're generating via Instaparse's transforms is runnable Clojure code, but we also have the opportunity to transform the arguments into any format we want. For instance, here, we could absolutely have the parser transform the key signature elements into a single map, or any other format we'd like.

A good REPL syntax might be something like:

    (key-signature {:e [:flat :flat] :c [:sharp]})

(which is equivalent to your idea -- I just used vectors instead of lists)

I like the idea of a hash-map as a single object that represents a key signature. Very easy to reason about. Conceptually, it makes sense, since you can't have multiple values for the same key, i.e., you can't have more than one list of accidentals for any given letter.

This is all stuff that I should be documenting in our `CONTRIBUTING.md`, but explaining it here is good practice, at least :P

src/alda/lisp/attributes.clj
daveyarwood: I have a feeling this problem might solve itself if you try hiding the `:pitch-notation` node like I mentioned. It would at least help us simplify things a bit, so we can figure out why exactly `:key-signature` is ending up in there. I'll bet it has something to do with the Instaparse transformation function for `:pitch-notation`

src/alda/lisp/model/pitch.clj
```diff
@@ -17,17 +17,38 @@
...
 (defn pitch
   "Returns a fn that will calculate the frequency in Hz, within the context
    of the octave that an instrument is in."
-  [letter & accidentals]
-  (fn [octave & {:keys [midi]}]
+  [parsed-pitch]
+  (let [letter (first parsed-pitch) accidentals (rest parsed-pitch)]
+  (fn [octave key-signature & {:keys [midi]}]
     (let [midi-note (reduce (fn [number accidental]
+                              (log/debug (format "number %s accidental %s" number accidental))
```
daveyarwood: I should add some guidance about `log/debug` statements to our `CONTRIBUTING.md`, but to keep a long story short, I think it would be good to remove this one before we merge the pull request.

I think the general rule should maybe be this: debug statements should generally be avoided, but they are OK if it's something that you would want to see when you turn "debug mode" on to troubleshoot a problem. Right now there are debug statements for each note event that is played -- logging out the pitch and duration, etc. -- and to log that certain components (e.g. the MIDI synth) have loaded. Things like that. If we're adding debug statements to help us understand what data is being passed around while we're implementing some new code, then the debug statements should probably be taken out.

(cc: @crisptrutski )

daveyarwood:
At the risk of discussing too many things in this thread, I do have one more idea:

What if you could use `key-signature` to define a "custom key" with any arbitrary flats and sharps that you want, and we set up a new attribute called just `key`, which would let you provide the *name* of a scale, instead?

So, these two attribute-setting statements would be equivalent:

    (key-signature {:f [:sharp] :c [:sharp] :g [:sharp]})

    (key :a :major)

@FragLegs if your head is about to explode, I'd be willing to step in and implement this part of the feature if you'd like :boom:

crisptrutski:
:+1: to the key *augmentation*

FragLegs:
That was going to be my next pull request! I'll roll it into this one
because it is a more natural interface for keys.

src/alda/parser.clj
FragLegs: So, would I then also add a `:key-signature` transform that looks like `#(apply parse-key-signature (apply parse-pitch %&))`?

daveyarwood: That should work -- I think it would be `#(parse-key-signature (apply parse-pitch %&))`

(`%&` is Clojure's special anonymous function syntax for "all the arguments, as a list," but the `parse-pitch` function takes a variable number of arguments (not a single list), so you can use `apply` to "unpack" the list. `parse-key-signature` takes a single list as an argument, so we don't need `apply` there)

FragLegs: Ohhh, I see. Because `<pitch-notation>` is now a hidden node, when I write `(key c++ a-)`, the parse-pitch function will get `('c' '+' '+' 'a' '-')`, whereas I was thinking it would get something more like `(('c' '+' '+') ('a' '-'))`.  I was confused about how each separate pitch notation got to `parse-key-signature`, but I think I get it now. Thanks!

daveyarwood:
@FragLegs Are you still working on this? I'm thinking of taking a look-see when I have a minute. Of note, we just updated master with an interesting feature (inline Clojure code - see #94) that will make the grammar/parsing part of this PR simpler. 

Basically, defining new attributes no longer requires adding anything to the grammar. Attributes are now implemented as built-in Clojure expressions that are evaluated in the context of the `alda.lisp` namespace, so all we should need to do is define the attribute via `defattribute`! (Plus whatever stuff needs to happen after the parsing stage)

FragLegs:
Awesome. I'll take a look when I can. My free time is pretty limited these
days though, so if you need this to move forward before I can get to it,
I'm fine with you picking it up.

daveyarwood:
Working on this now -- going pretty well so far.

I just realized that there is already a [`key` function in clojure.core](https://clojuredocs.org/clojure.core/key), so if we want to create a separate attribute for specifying a key signature like `:c :major`, it will have to be called something besides `key`... any ideas? Off hand, I can't think of what I would call the "X" a song is in, if not the "key"... hmm...

FragLegs:
I was thinking of leaving it under key-signature, but allowing the user to
add capital letters as shorthand. That is (key Am) would give you the A
minor key and (key a- b-) would give you a key with a flat a and flat b.

crisptrutski:
@daveyarwood I think calling the function `key` if fine, and we should find a solution to namespace clashes.

The "try alda.lisp namespace first, then fallback to current namespace" has some issues, to furnish an example:

```clojure
(test-parse :clj-expr "(let [tempo 1] (+ 1 tempo))")
=> (clojure.core/let [alda.lisp/tempo 1] (clojure.core/+ 1 alda.lisp/tempo))
```

So perhaps a better approach would be to set up a "clean room" namespace, with the appropriate `alda.lisp` functions referenced, and the conflicting `clojure.core` functions removed.

daveyarwood:
I'm hesitant to overwrite any clojure.core functions... someone may want to use `clojure.core/key`, for instance. I avoided calling the function that creates a musical "rest" `rest` because of `clojure.core/rest`, for example. I'm sure we can think of another name for "key." Maybe `scale`?

daveyarwood:
Snowballing off of @FragLegs' idea, what if we used `key-signature` (or its abbreviation `key-sig`) for both kinds of key-setting, and intelligently did the right thing depending on the type of the argument?

i.e. you could set the key to A major in one of three ways:

```clojure
(key-sig [:a :major])
(key-sig {:f [:sharp] :c [:sharp] :g [:sharp]})
(key-sig "f+ c+ g+")
```

daveyarwood:
I've copied this to a branch and made some tweaks -- see PR #110

PR #90, opened 2015-09-13
Reloadable REPL

crisptrutski:
Having a REPL that live-loads changes made to the source base, without depending on piggy-backing on a Clojure REPL, would be exciting and accessible for new users. It would also be invaluable with the server.

It could be as simple as `boot watch alda -x repl`, and there could be hash-bang sugar for that.

So far this is just the first tiny step.. having the alda task run via the regular boot lifecycle instead of launching statically (haven't measured overhead in start-up time, hopefully negligible). The benefit of this is that the REPL will automatically relaunch with changes.. but you have to manually close the old REPL first. On the plus-side the relaunch is basically instantaneous.

Now for the bad news sandwich - the relaunch crashes: `javax.sound.midi.InvalidMidiDataException: cannot get soundbank from stream`

Haven't spent much time on this, might be pretty low hanging. 

Some ideas to take this forward:

- [x] Don't crash from synth contention (blacklisted namespace? run this bit in separate singleton pod?)
- [ ] Don't block in the task - catch code changes immediately
- [ ] Refresh currently running REPL, and give some on-screen feedback (success / warning / fail)
- [ ] Don't actually relaunch on kill, in fact kill the `watch`-er to save double-quit
- [ ] Don't lose any state (perhaps by blacklisting namespaces for reload, or serializing and restoring it)

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -19,7 +19,7 @@
          '[alda.cli]
          '[str-to-argv :refer (split-args)])
 
-; version number is stored in alda.version 
+; version number is stored in alda.version
 (bootlaces! alda.version/-version-)
 
 (task-options!
@@ -50,11 +50,17 @@
 
 (deftask alda
   "Run Alda CLI tasks.
-   
-   Whereas running `bin/alda <cmd> <args>` will use the latest deployed 
+
+   Whereas running `bin/alda <cmd> <args>` will use the latest deployed
    version of Alda, running this task (`boot alda -x '<cmd> <args>'`)
    will use the current (local) version of this repo."
   [x execute ARGS str "The Alda CLI task and args as a single string."]
-  (when execute
-    (let [cli-args (split-args execute)]
-      (apply (resolve 'alda.cli/-main) cli-args))))
+  (fn [next-task]
+    (fn [fileset]
+      (require '[alda.version] :reload-all)
+      (require '[alda.repl] :reload-all)
+      ;(require '[alda.cli] :reload-all)
+      (when execute
+        (let [cli-args (split-args execute)]
+          (apply (resolve 'alda.cli/-main) cli-args)))
+      (next-task fileset))))


daveyarwood:
> Now for the bad news sandwich - the relaunch crashes: `javax.sound.midi.InvalidMidiDataException: cannot get soundbank from stream`

Someone reported seeing this exception in the context of running `(alda.cli/-main "play" "--file" "/path/to/some/file")` more than once from a Boot REPL. I played around with it a little and I think the issue is related to trying to re-load the soundfont into an existing MIDI synth where it was already loaded. Including the "--stock" flag fixes it, although of course we would want to support having the soundfont loaded every time.

daveyarwood:
Once we're at 1.0.0 with @jeluard's PR #107 merged in, the "cannot get soundbank from stream" exception ought to be resolved, at least. :+1:

daveyarwood:
Things are a lot different now that the new server/client structure has hit master. We no longer have a `boot alda` task, and the Alda REPL will soon be rewritten as part of the Java client.

For development on the Alda REPL, I think you could still do something in the spirit of this PR, only the entrypoint would be `alda.repl/start-repl!` instead of `alda.cli/-main` (`alda.cli` is no more). I think it's kind of futile though, with the Alda REPL about to be ported to the client.

@crisptrutski: Give the new server/client stuff in 1.0.0b a whirl, if you get a minute. I'm curious to know what you think, and see if you have any ideas for how to optimize Alda development. So far, my development cycle has been:

* Make some changes to either the client (Java) or server (Clojure) code
* Run `boot build -o /tmp` (you could automate this by using the `watch` task)
* If there were server changes, run `/tmp/alda restart` to restart the server
* If there were client changes, no need to do anything -- they're in `/tmp/alda`
* Try out the changes
  * Running `/tmp/alda play -c '<some code here>'` is just like typing lines into the Alda REPL
  * The client sends the code snippets to the Alda server, which handles them and plays the result

Seems like we could probably add a Boot task that watches for changes and refreshes a development Alda server running in the foreground. Maybe via Ring `wrap-reload`?

daveyarwood:
Closing this PR for now to keep things tidy -- we can always revisit in the future.

PR #95, opened 2015-09-18
Include bar-lines in the parse tree

daveyarwood:
This is a change that will help facilitate visually rendering scores. Bar lines are now parsed as a [:barline] node, which is transformed to a call to (alda.lisp/barline), which does NOTHING. IT DOES ABSOLUTELY NOTHING.


daveyarwood:
:rice_cracker: :abc: :honeybee: :briefcase: :boy:

PR #102, opened 2015-09-23
Inline Clojure code

### LINKED ISSUE #94: Inline Clojure code
This feature idea has come up for discussion in various other issues, where it became apparent that we could actually implement other features in terms of inline Clojure code. Example: #87 

Quoting myself from that issue:

> Reserving parentheses for this in the Alda grammar would be pretty badass IMO. Currently the only thing using them is attribute-change events, but maybe we could hack in a way to re-implement those in terms of embedded Clojure code?

Moving toward doing as many things as possible with inline, embedded Clojure code would carry the benefit of not polluting the syntax of Alda with too many different symbols. For backwards compatibility, we should keep the symbols we have (e.g. `%marker`, `@at-marker`), but with inline Clojure code, we should also be able to use these features via the alda.lisp namespace, i.e. `(marker "foo")`, `(at-marker "bar")`, which some users might even prefer. We'll just make sure that we document that these things can be done in more than one way.

I think the set of symbols that we're currently using are for common-enough tasks that it makes sense that they also get the privilege of having their own first-class syntax in the Alda grammar, but as we add new features to Alda (microtonal music, etc.), it will only get harder and harder to find symbols that we haven't used yet, so I think this is a step in the right direction.

---

EDIT: 
Checklist:

- [x] Grammar/parser changes: read anything between parens as a Clojure expression
- [x] Transform Clojure expressions like `[:clj-expr "(+ 1 1)"]` into i.e. `'(+ 1 1)`
- [x] macroexpand commas and semicolons into separate expressions, i.e. `(volume 50, quant 50)` becomes `(do (volume 50) (quant 50))`
    - [x] commas/semicolons can exist in strings...
    - [x] ...and character literals (`\,` and `\;`)
- [x] Custom namespace resolution -- try to resolve in the context of alda.lisp first, and if that fails, evaluate in the current namespace at runtime (whatever that is)
- [x] Make sure `defattribute` is doing all the things we need it to do.
    - [x] Define bang-suffixed versions of the setter fns that do a global attribute change.
    - [x] Make sure it defines setter functions for all aliases of the function -- I don't think it does.

daveyarwood:
Closes #94.

This introduces a major change to `alda.parser` -- anything between parentheses is now interpreted as a Clojure expression and evaluated with the help of a custom reader that will attempt to resolve symbols in the `alda.lisp` namespace whenever possible. There is also some regex hackery that allows things like `(volume 50, tempo 120)` to still work by splitting on commas and semicolons (intelligently\*) and turning them into separate S-expressions like `(do (volume 50) (tempo 120))`.

\*Commas and semicolons inside strings and as character literals still function as they normally do in Clojure.

As a side effect of this, `(* the old (and not so great) long comment syntax *)` had to be deprecated. R.I.P.

For full backwards compatibility, I've also extended the `defattribute` macro so that it defines more functions for use in the `alda.lisp` namespace (and by extension, in Alda code) -- you can now call any alias of an attribute as a function (e.g. `(vol 50)` instead of having to do `(set-attribute :vol 50)`), and you can set global attributes by appending a `!` after the function name (e.g. `(vol! 50)` instead of `(global-attribute :vol 50)`)

---

I also happened to notice a bug with the `:play` function not fully resetting instruments' attributes (octave, volume, etc.) before playing a score from the beginning, leading to inconsistent playback. I'm including a fix for that in this PR as well.


daveyarwood:
cc: @crisptrutski 

I'm feeling pretty good about this -- looping you in in case you have any thoughts on this before I merge it

crisptrutski:
Very exciting! More than happy to review, will have some time tomorrow

pkobrien:
This is going to be very interesting. I've only read the discussions and haven't had a chance to review the code, but I really like the direction you've taken. I think this will open up some really clever and powerful possibilities. Very nice.

src/alda/lisp/attributes.clj
```diff
@@ -56,7 +44,6 @@
    have a duration of 450 ms, and the next event will be set to occur in 500 ms."
   :aliases [:quant :quantize]
   :initial-val 0.9
-  :fn-name quant
```
daveyarwood: Deleting this line, as the changes to `defattribute` make it so that functions are defined for `quantization`, `quant` and `quantize` based on the attribute name and aliases. (Note: If I had left it in, an alias would not be defined for `quantization`)

daveyarwood:
Reviews welcome! :dolphin:

crisptrutski:
Very slick change, seems to well!

The regex voodoo is unfortunately a bit too naive in my opinion, consider a silly example like ` (prn [1,2,3])`. The current treatment will result in unbalanced parens for separators anywhere inside a nested expression.

My suggestion would be to "use the grammar", Luke. Perhaps something like `clj-expr-with-separators  =  <"("> inside-clj-expr (<special-clj-separator> inside-clj-expr)* <")"> <ows>
`

Not sure if this is related to the `:play` reset change, but getting odd behaviour now in the REPL - there's a long pause before playback of the new notes starts, and also the durations are not reliable.

Could very well be something goofy on this machine of mine, recently got new hardware and went through the time machine warp gate.

daveyarwood:
@crisptrutski My initial thought was that we could prescribe that people not use Clojure "whitespace commas" in embedded Clojure code within Alda, but it would be nice if we could properly handle things like `(prn [1, 2, 3])` and `(prn {:a 1, :b 2})`. I tried adding square brackets and curly braces to the regex, but it soon became apparent that the level of complexity of this regex is starting to spiral out of control. The regex would need to account for "commas/semicolons inside square brackets, but not if the square brackets are inside of a string", etc. :persevere: 

We should use a parser-based approach instead, for sure. Unfortunately, I don't think we can add this to the existing Alda grammar, because the `clj-expr` rule matches nested Clojure expressions by reading one character (OR another `clj-expr`) at a time, stopping once it gets to a `)`. I experimented with this a lot when I was trying to implement long comments, and it seems like there is no possible way to end up with a single string/parse-node containing all of the Clojure code, e.g.:

    [:clj-expr "+ 1 1 " [:clj-expr "* 2 3"]]

Instead, what we get is:

    [:clj-expr "+" " " "1" " " "1" " " [:clj-expr "*" " " "2" " " "3"]]

As far as I can tell (correct me if I'm wrong), we will have to parse the Alda score with `alda.bnf`, then transform each `clj-expr` node into an alda.lispified custom Clojure expression using a separate parser, i.e. `clj_expr.bnf`. I don't like this approach, but I guess it's not terrible. I can't really think of any problems with it.

I also found another problem -- the `clj-expr` rule does not account for open/close paren character literals or parentheses inside a string. Arrgh! I guess I could copy the regex voodoo into alda.bnf (twice, one for the open paren and one for the close paren), but there has to be a better way. Any ideas?

I'll add a few test cases and take a closer look later -- but if anyone is feeling intrepid, feel free to jump in :surfer:

daveyarwood:
@crisptrutski Not sure what's up with the audio playback issues... I tested this branch on the example scores and in the REPL and I didn't notice any problems.

crisptrutski:
@daveyarwood Oh, that does sound tricky. Going to try something closer to what you have right now, but using `clojure.core/read-string`

crisptrutski:
D'oh, was expecting the reader to keep commas as symbols. That was silly of me :disappointed:

daveyarwood:
If all else fails, we could just give up on the idea of splitting on commas and semicolons, and prescribe that people do one of these instead of `(volume 50, quant 100)`:

    (do (volume 50) (quant 100))

    (volume 50) (quant 100)

I like the `(volume 50, quant 100)` syntax because it blurs the line between writing Clojure code and "creating a multiple-attribute-changes event" in the eyes of the user, and it looks nice.

I'm on the fence, though. Part of me thinks that it might put Clojure programmers' minds more at ease if the inline Clojure code in Alda works just like writing Clojure code, including commas-as-whitespace. This could lead to less headaches if, for example, you're copy-pasting Clojure code into an Alda file and it happens to use commas as whitespace within an S-expression.

crisptrutski:
Personally I like consecutive s-exprs quite alright, and you could always add a function just sugar setting attributes (say `(multi volume 50, quant 100)`, where `,` is whitespace again and multi is some kind of `(do ~@(partition 2 body))` macro.

A crude, but seems-to-work hack on top of `read-string` is [here](https://github.com/alda-lang/alda/tree/injurious-clojure), that just needs changes for grammar to cope. Have a suspicious that we can still make this problem easier with grammar changes though. 

Another awful idea is to to see if we can abuse `tools.reader` to our will, but guessing that would require monkey spanners or forking.

daveyarwood:
Crude as it may be, it's still way better than resorting to voodoo! :stuck_out_tongue_winking_eye: 

This seems to have done the trick, and is way easier to read than that crazy regex. Spot on! :boom:

I think the only thing that needs fixing now is the way the grammar interprets anything inside parens as an S-expression, right? I think that might be fixable by defining rules in the grammar to match strings and character literals, and amending the `clj-expr` rule to pick up on these atomically while it's reading one character at a time.

daveyarwood:
Alright, I've got it working! :fireworks: :dolphin: :guitar: :notes:

I'm working on iterating further on this. I bet if I add square brackets and curly braces to the grammar, I can reduce the hackiness of what's going on in alda.parser

daveyarwood:
I just went off the deep end a little, check out that last commit.

In the process of trying to make the code simpler, I made it more complicated, but it was out of necessity -- this version is more robust and it catches more edge cases (see the new tests I added). 

The grammar is doing a little more of the work this time in that it can recognize Clojure S-expressions, strings, character literals, vectors, maps and sets. But there is also a fairly complex transformation that occurs on each `clj-expr` node. Basically, it loops through every character of text that wasn't parsed as one of the other types (string, character literal, vector, etc.) and adds it onto the "final" string to be read as alda.lisp code. If it encounters a comma or semicolon in the process, it starts building another string from scratch starting with the next character. Clojure character and string literals are treated specially by wrapping them in a "StringHack" record -- this is so that the `clj-expr` node knows to ignore their contents in case they contain commas or semicolons which should not be split on.

Some of this code may be a little hacky or not as performant as it could be -- let me know if you see any areas where it could be refactored or improved. Otherwise, I'm pretty happy with it.

crisptrutski:
Have only skimmed, but this looks like the winning approach when you consider its predecessors :+1:

daveyarwood:
@crisptrutski Would you like to give this a proper review, or shall I go ahead and merge it? I'm fine either way

crisptrutski:
No need to tarry on my behalf!

daveyarwood:
:metal:

pkobrien:
I have to say I'm not a fan of the inline commas idea. If it was me I would drop support for them as I fear they are going to end up being a source of trouble and complexity that could be avoided.

daveyarwood:
@pkobrien I think that's a valid concern. Personally, I could go either way on this, and am not necessarily tied to the comma idea. We should discuss this a little more. It sounds like so far, no one else has spoken in favor of splitting expressions on commas and semicolons, so maybe we should axe that idea after all.

pkobrien:
I just think you'll be saving yourself a lot of time and effort in the long run if your language doesn't include anything that violates Clojure syntax. As @crisptrutski has pointed out you could get some of what you were after by using a macro. This is your project and your vision and your decision. I just thought I'd throw out my opinion while you are in the early stages here. You might want to think this one over before claiming it as a feature. It will be hard to remove later, but easy to add it back in if you still feel good about it. Just my 2 cents. :-)

daveyarwood:
I think we should go ahead and remove it for now -- I don't like the idea of breaking existing scores that might happen to use the comma syntax (knowingly or not), but perhaps more importantly, the idea of Clojure code not evaluating like it usually would just doesn't sit well with me, and could very well lead to problems and confusion down the line. We could always add it back in in the future if someone makes a good case for it.

I'll create a new issue to simplify the inline Clojure code -- I'll probably get to it within the next few days if someone else doesn't grab it first :)

daveyarwood:
Well that was easy.

PR: #109

PR #106, opened 2015-09-26
Simplify CLJ expression parser

crisptrutski:
Looks like it works, but perhaps too naive? Certainly seems simpler :wink: 

Watch out, this contains #105. For a clean diff just look at e41ca1c8c7473052e2eaeb90a2b2492dc28b461e


daveyarwood:
Whoops! Good catches. This seems to do the trick! I'm gonna merge it. Thanks!

PR #107, opened 2015-09-27
Replaced existing FluidR3 installation mechanism with a dedicated script

jeluard:
Moving to a script that installs FluidR3 for the local user has to be done only once and does not impact `alda` startup. It also allows to start without internet access.

Having such a default soundbank removes the whole process of loading an extra soundbank, removing existing instruments and loading new ones each time `alda` runs.

Finally it is a step in the direction of removing `boot` dependency from `alda`.

(Sorry for the whitespace changes, somehow my editor has been too zealous).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing to Alda
 
-Pull requests are warmly welcomed. Please feel free to take on whatever [issue](https://github.com/alda-lang/alda/issues) interests you. 
+Pull requests are warmly welcomed. Please feel free to take on whatever [issue](https://github.com/alda-lang/alda/issues) interests you.
 
 ## Instructions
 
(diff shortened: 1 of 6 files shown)


jeluard:
Installing in the local `.gervill` folder looks the best option.

You can see [here](http://www.docjar.com/html/api/com/sun/media/sound/SoftSynthesizer.java.html) (starting line 625) the complete logic to load a soundfont file. Before loading the local `soundbank-emg.sf2` (emg stands for emergency) a per JVM file is loaded. This approach unfortunately appears to only work with JDK and forces user to repeat the same operation for every JVM upgrade.

Also please note that when removed this file is regenerated by the JVM using default instruments so it's safe to overwrite.

daveyarwood:
This is awesome - thank you! I tried it locally, and it works great.

I think instead of merging this to master right away, we should start a `1.0.0` branch that will contain big changes like this\*, which will make all of the changes more noticeable once alda 1.0.0 is officially released. @crisptrutski @pkobrien @alandipert  Thoughts on this?

\*(Running Alda via an uberjar wrapped in a binary executable will be another such change.)

jeluard:
Makes sense! Do I have to do something to merge to branch `1.0.0`? This is where my git-fu stops.

Let me know if you want to pursue this approach to reduce startup time and I will continue sending patches.

daveyarwood:
I don't think there is a way to merge a PR to a different branch from the GitHub UI, as far as I know... don't worry, I'll take care of it :)

I'm very interested in the approach we've been talking about -- please do continue to send patches! :+1:

daveyarwood:
I've pushed a `1.0.0` branch.

PR #109, opened 2015-09-28
Simplify inline Clojure code

### LINKED ISSUE #108: Simplify inline Clojure code
Per the discussion in #102, we are deciding to throw out my initial idea of making attributes backwards-compatible by splitting S-expressions on commas/semicolons. We can always add it back in in the future if someone presents a good case for it, but for now it seems like it is a better idea for our inline Clojure code to be consistent with regular ol' Clojure code, where you wouldn't expect whitespace commas to potentially do anything weird.

This will be a breaking change for scores that use the comma syntax (under the guise of a "multiple attribute change" under the old grammar), so we should push this out as a new minor version number.

daveyarwood:
Closes #108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # CHANGELOG
 
+## 0.?.0 (...)
+
+* Simplify inline Clojure expressions -- now they're just like regular Clojure expressions. No monkey business around splitting on commas and semicolons.
+
+### Breaking changes
+
+* This breaks backwards compatibility with "multiple attribute changes," i.e.:
+
+        (volume 50, tempo 100)
+
+    This will now attempt to be read as a Clojure expression `(volume 50 tempo 100)` (since commas are whitespace in Clojure), which will fail because the `volume` function expects only one argument.
+
+    To update your scores that contain this syntax, change the above to:
+
+        (do (volume 50) (tempo 100))
+
+    or just:
+
+        (volume 50) (tempo 100)
+
+---
+
 ## 0.9.0 (10/1/15)
 
 * Implemented panning via the `panning` attribute.
--- a/bin/alda
+++ b/bin/alda
@@ -9,7 +9,8 @@
 
 (set-env! :dependencies '[[alda "LATEST"]])
 
-(require '[alda.cli])
+(require '[alda.cli]
+         '[alda.lisp :refer :all])
 
 (defn -main [& args]
   (apply (resolve 'alda.cli/-main) args))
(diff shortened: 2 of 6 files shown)


crisptrutski:
After all my hard work simplifying things :wink:

daveyarwood:
Sorry @crisptrutski! :)

daveyarwood:
Quoth @crisptrutski in #89:

> The "try alda.lisp namespace first, then fallback to current namespace" has some issues, to furnish an example:
>
> ```clojure
(test-parse :clj-expr "(let [tempo 1] (+ 1 tempo))")
=> (clojure.core/let [alda.lisp/tempo 1] (clojure.core/+ 1 alda.lisp/tempo))
```
>
> So perhaps a better approach would be to set up a "clean room" namespace, with the appropriate `alda.lisp` functions referenced, and the conflicting `clojure.core` functions as exclusions.

We should probably fix this before merging this PR -- seems like it could lead to unexpected behavior :/

daveyarwood:
Hmm... this is a tricky problem. The Clojure reader must be able to differentiate somehow between symbols in bindings (and their subsequent references) and symbols elsewhere, so that it doesn't try to resolve them, but I haven't figured out how.

daveyarwood:
@crisptrutski Check this out -- I talked to @micha, and it turns out that all we really need to do to get this working is to refer `alda.lisp` into the `boot.user` context. This leverages the way that Boot creates a script at run-time and then evaluates it (you can see the script by running `boot -b`) in the `boot.user` namespace.

I've added `(require '[alda.lisp :refer :all])` to both the `alda` script and the `build.boot`, so that this will work both when running the script and the `boot alda` task.

This will get trickier in 1.0.0 when we're packaging Alda as an uberjar, and thus won't have access to `boot.user`, but I think this will do for now.

crisptrutski:
Looks great, this is exactly what I was thinking also :+1:

daveyarwood:
Brandon Bloom (creator of backtick) had this to say:

> Backtick is an implementation of syntax-quote, which operates at the reader level. While Clojure's syntax-quote does leverage knowledge of the dynamic environment (namespaces, vars, etc), it does not have any notion of lexical scope or any high-level understanding of Clojure forms. When your templating code, the normal advice for writing macros applies: Use gensyms:

>     `(let [tempo# 1] (+ 1 tempo#))

> But really, you probably don't want syntax-quote or backtick at all for this use case. Instead, use the normal reader, wrap with `#(list 'fn [] %) and use a combination of in-ns/use/eval.

That last idea -- wrapping the expressions in an anonymous function and using `in-ns` and/or `use` in combination with `eval` -- is worth a try.

crisptrutski:
Reminds me a bit of JavaScript :wink:

Issue #104, opened 2015-09-25
Provide superpowers for inline Clojure code


daveyarwood:
Now that inline Clojure code is implemented, it would be useful to have a couple of new things defined in `alda.lisp`, ready for use in inline Clojure code:

* An `alda-code` function or macro that takes a string as an argument and parses it at eval-time. This would be super sweet because it would allow users to generate strings of Alda code using inline Clojure code, and then have them be evaluated as if they were part of the score.
    * I'm not sure this would work 100% of the time (since `clj-expr` nodes can only be placed in certain positions according to the Alda grammar), but maybe we don't necessarily need it to.
    * simple example: `c d e (alda-code "f") g` would evaluate to:

            (alda.lisp/note (alda.lisp/pitch :c))
            (alda.lisp/note (alda.lisp/pitch :d))
            (alda.lisp/note (alda.lisp/pitch :e))
            (alda.lisp/alda-code "f")
            (alda.lisp/note (alda.lisp/pitch :g))

        and then `(alda.lisp/alda-code "f")`, when evaluated, would parse `f` into `(alda.lisp/note (alda.lisp/pitch :f))` and evaluate it, effectively doing the same thing as just writing `f` in your score instead of `(alda-code "f")`. This is a dumb example because why wouldn't you just write `f`? But the power lies in being able to use Clojure to dynamically generate strings of Alda code, *at runtime, inside your Alda score*.

* A way to schedule function call events the way that note events are scheduled. Right now, if you insert `(println "hey there!")` in the middle of an Alda score, the printing will happen right away when the score is evaluated. But wouldn't it be cool if we could add some sort of event-scheduling function to `alda.lisp`? I'm envisioning being able to print out things at certain points in the score, or basically do *anything* at any point during the playback of the score, by simply adding something like `(schedule #(println "drop the bass"))` right where you want it to happen in the score. The `schedule` function would only need to add the function to the score's `:events` as a sort of "function event" that includes an offset, which would be determined in the same way as the offset of note events. Then we could teach `play-event!` how to handle these function events. Easy peasy.

daveyarwood:
I've started implementing `alda-code` on a branch, for use as a stepping stone to implementing repeats :guitar: :notes:

PR #113, opened 2015-10-02
Code block literals and the "alda-code" function

daveyarwood:
This PR implements the `alda-code` function discussed in #104, as well as code block literals, as a stepping stone to implementing repeats (#7).

Code blocks are captured in the grammar as anything between `[ square brackets ]`. At the moment, they are essentially ignored when the score is evaluated -- they simply evaluate to strings. 

I plan to implement repeats as an alda.lisp function `repeat-code` that will take a string of text, repeat it a given number of times, and evaluate it using `alda-code`. There will be some syntactic sugar in the form of `xN` (where `N` is the number of times to repeat) and/or `||:  code here  :||N` which will be parsed as a call to `repeat-code`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ of music: classical, popular, chiptune, electroacoustic, and more!
 * Represent scores as text files and play them back with the `alda` command-line tool
 * [Interactive REPL](doc/alda-repl.md) lets you type Alda code and hear the results in real time
 * [Underlying Clojure DSL](doc/alda-lisp.md) allows you to [use Alda directly in your Clojure project](doc/alda-now.md).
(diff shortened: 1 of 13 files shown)


daveyarwood:
REPL seems to hang when I do `piano: [ c d e ] f` (correctly plays just the F), then `[c]` (hangs)

crisptrutski:
My guess is that hang was due to the bug fixed in aeee9c656b51ba0ac1c14c77592ba99c20016ab4 - playing 0 notes

daveyarwood:
@crisptrutski Yep -- problem solved!

Will merge & deploy this in the morning. On to repeats next!

PR #117, opened 2015-10-04
Implement [C]razy [R]hythm [A]lda [M]agic

### LINKED ISSUE #9: Implement advanced rhythms
**Cram**

I came up with this idea for representing advanced rhythms. I call it "cram" because you're basically cramming a bunch of notes inside of a set duration and dividing them up either evenly or according to a weight distribution. 

(**EDIT 9/7/15:** @crisptrutski has a much better idea -- see comments below)
Desired syntax:

    { d+ c+ < b }2 

The above is one way to represent a quarter note triplet -- it's 3 notes crammed into the space of a half note. By default, each note inside the braces is equal in length.

Notes inside of "cram brackets" use a slightly different syntax than usual, in that you don't specify exact note durations like eighth, quarter, etc. Instead, you can (optionally) specify how much weight each crammed note gets, by adding \*'s to the notes you want to be relatively longer than the others. This can get rather complicated, but here is an simple example:

    { c* d }4

The "c" has one \*, and the "d" has none, and they are crammed into the space of a quarter note (4). By default, a note has no \*, which corresponds to a ratio of 1:1 (i.e. all notes are equal in length). With each \* that you add, the first number of the ratio increases by one. So what this means is that in the example above, the "c" note (c*) has a length ratio of 2:1, whereas the "d" note (d) is just 1:1 (technically 1:2, compared to the "c" note). Because the "c" note is twice as long as the "d" note, and the two notes are crammed into the space of a quarter note, the result is two "swung" eighths, or in other words, an eighth note triplet where the "c" occupies the first two notes, and the "d" occupies the third.

Perhaps an easier way to think about adding \*'s to notes within cram brackets, is that with each \* you add, it's like adding another note into the crammed group, only instead of a new note, you're repeating the note you *'d, and then tying them together.

It should be possible to include rests and chords inside of cram brackets, too.

**Second/millisecond durations**

Come up with some syntax for overriding the beat/tempo model and letting users define exact durations of notes in seconds and milliseconds. Could be as simple as `c60s` (a C lasting 60 seconds), `d+500ms` (a D# lasting 500 milliseconds), etc.

crisptrutski:
Closes #9

### DIFF
--- a/src/alda/lisp/cram.clj
+++ b/src/alda/lisp/cram.clj
@@ -1,0 +1,40 @@
+(ns alda.lisp.cram)
+
+(in-ns 'alda.lisp)
+
+(def no-events
+  {:start {:offset (->AbsoluteOffset 0), :events []}})
+
+(defmacro resetting [vars & body]
+  (if (seq vars)
+    (let [[x & xs] vars]
+      `(let [before# ~x
+             result# (resetting ~xs ~@body)]
+         (alter-var-root (var ~x) (constantly before#))
+         result#))
+    `(do ~@body)))
+
+(defmacro calc-duration [& body]
+  `(resetting
+    [~'alda.lisp/*events*
+     ~'alda.lisp/*instruments*
+     ~'alda.lisp/*current-instruments*
+     ~'alda.lisp/*global-attributes*]
+    (let [start#  (:offset ($current-offset))]
+      (alter-var-root ~'#'alda.lisp/*global-attributes* #(dissoc % :duration))
+      ~@body
+      (- (:offset ($current-offset)) start#))))
+
+(defmacro cram [& body]
+  (let [lst (last body)
+        dur (and (list? lst)
+                 (= 'alda.lisp/duration (first lst))
+                 lst)
+        body (if dur (butlast body) body)]
+    `(let [b#       (or ~dur ($duration))
+           d#       (calc-duration ~@body)
+           t#       ((:duration-fn (duration b#)) ($tempo))
+           scaling# (* (/ t# d#) ~'alda.lisp/*time-scaling*)]
+       (binding [~'alda.lisp/*time-scaling* scaling#]
+         ~@body
+         (set-duration b#)))))
--- a/src/alda/lisp/cram.clj
+++ b/src/alda/lisp/cram.clj
@@ -21,7 +21,6 @@
      ~'alda.lisp/*current-instruments*
      ~'alda.lisp/*global-attributes*]
     (let [start#  (:offset ($current-offset))]
-      (alter-var-root ~'#'alda.lisp/*global-attributes* #(dissoc % :duration))
       ~@body
       (- (:offset ($current-offset)) start#))))
 
--- a/src/alda/lisp/cram.clj
+++ b/src/alda/lisp/cram.clj
@@ -21,16 +21,17 @@
      ~'alda.lisp/*current-instruments*
      ~'alda.lisp/*global-attributes*]
     (let [start#  (:offset ($current-offset))]
+      (set-duration 1)
       ~@body
       (- (:offset ($current-offset)) start#))))
 
 (defmacro cram [& body]
   (let [lst (last body)
-        dur (and (list? lst)
+        dur (and (or (coll? lst))
                  (= 'alda.lisp/duration (first lst))
                  lst)
         body (if dur (butlast body) body)]
-    `(let [b#       (or ~dur ($duration))
+    `(let [b#       (if ~dur (:beats ~dur) ($duration))
            d#       (calc-duration ~@body)
            t#       ((:duration-fn (duration b#)) ($tempo))
            scaling# (* (/ t# d#) ~'alda.lisp/*time-scaling*)]
--- a/grammar/alda.bnf
+++ b/grammar/alda.bnf
@@ -15,7 +15,7 @@ nickname                = <"\""> acceptable-name <"\"">
 <music-data>            = <ows> (voices | event)*
 <event>                 = chord | note | rest | octave-change |
                           clj-expr | marker | at-marker | barline |
-                          code-block
+                          code-block | cram
 voices                  = voice+ (<voice-zero> | <#"\z"> | &part)
 voice                   = voice-number event*
 voice-number            = <"V"> #"[1-9]\d*" <":"> <ows>
@@ -70,6 +70,10 @@ clj-expr                = <"("> inside-clj-expr* <")"> <ows>
 code-block              = code-block-no-ws <ows>
 code-block-no-ws        = <"["> inside-code-block* <"]">
 <inside-code-block>     = !( "[" | "]") #".|\n|\r" | code-block-no-ws
+
+    (* cram - rhythmic section *)
+
+cram                    = <"{"> event* <"}"> duration? <ows>
 
     (* comments & whitespace *)
 
--- a/src/alda/lisp.clj
+++ b/src/alda/lisp.clj
@@ -7,8 +7,7 @@
             [alda.lisp.events]
             [alda.lisp.instruments]
             [alda.lisp.score]
-            [alda.lisp.code]
-            [alda.lisp.cram]))
+            [alda.lisp.code]))
 
 ; example: creating a score interactively at the repl
 
--- a/src/alda/lisp.clj
+++ b/src/alda/lisp.clj
@@ -7,7 +7,8 @@
             [alda.lisp.events]
             [alda.lisp.instruments]
             [alda.lisp.score]
-            [alda.lisp.code]))
+            [alda.lisp.code]
+            [alda.lisp.cram]))
 
 ; example: creating a score interactively at the repl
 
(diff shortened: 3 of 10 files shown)


crisptrutski:
So here's a step towards a "execute, rollback, push-dynamic-scope, re-execute" approach - not the most efficient or elegant, but quite close to working.

Remaining work is to:

1. Calculate correct `*time-scaling*`
2. Apply `*time-scaling*` within `note*`
3. Make sure the appropriate state is set after (notably duration)

Think I've got a handle on how to do these, continuing on

crisptrutski:
@daveyarwood is it possible that multiple instruments would play the same cram at the same time, but with different durations? wondering if I need to calculate and set `*time-scalings* per instrument.. (eish)

daveyarwood:
@crisptrutski It should be a per-instrument thing for sure. You could implement it as an attribute via `defattribute`

daveyarwood:
@crisptrutski Actually, scratch that -- it doesn't need to be an attribute. I think `*time-scaling*` can still be a global dynamic var, it should just be used to calculate things differently per-instrument, depending on that instrument's duration.

crisptrutski:
Hope you're right, but I think the issue is that both instruments will see the same literals "inside" the cram, which override their currents duration attribute. An instrument playing at half the meter of another one would need to scale those notes twice as sharply.

Haven't really played around with multiple instrument parts, so this gives me an excuse to test it out :surfer:

src/alda/lisp/cram.clj
```diff
@@ -0,0 +1,41 @@
...
+         (resetting ~xs ~@body)))))
+
+
+(defmacro calc-duration [& body]
+  `(resetting
+    [~'alda.lisp/*events*
+     ~'alda.lisp/*instruments*
+     ~'alda.lisp/*global-attributes*]
+    (let [data# (do (alter-var-root ~'#'alda.lisp/*events* (constantly ~alda.lisp/no-events))
+                    (alter-var-root ~'#'alda.lisp/*global-attributes* #(dissoc % :duration))
```
daveyarwood: I'm not sure this is necessary -- `*global-attributes*` is actually a map of offsets to attribute change events.

To illustrate how it works:

```
alda.lisp=> (score*)
{}
alda.lisp=> (tempo! 90)
#alda.lisp.GlobalAttribute{:offset #alda.lisp.AbsoluteOffset{:offset 0}, :attr :tempo, :val 90}
alda.lisp=> (part* "piano")
nil
alda.lisp=> (do (note (pitch :c)) (note (pitch :d)) (note (pitch :e)))
(#alda.lisp.Note{:offset #alda.lisp.AbsoluteOffset{:offset 1333.3333740234375}, :instrument "piano-m7CA6", :volume 1.0, :track-volume 0.7874015748031497, :panning 0.5, :midi-note 64, :pitch 329.6275569128699, :duration 600.0000183105469})
alda.lisp=> (volume! 80)
#alda.lisp.GlobalAttribute{:offset #alda.lisp.AbsoluteOffset{:offset 2000.0000610351562}, :attr :volume, :val 80}
alda.lisp=> *global-attributes*
{0 [[:tempo 90]], 2000.0000610351562 [[:volume 80]]}
```

daveyarwood:
> Hope you're right, but I think the issue is that both instruments will see the same literals "inside" the cram, which override their currents duration attribute. An instrument playing at half the meter of another one would need to scale those notes twice as sharply.

I don't think that's the behavior that we want, though. If there are two current instruments, playing at two different tempos, then we would want the crammed notes to be shorter for the faster instrument. I haven't thought too deeply about the math behind this yet, but I would think the scale factor would still be the same, right? The scale factor would be applied to the duration (so it would be the same for both instruments), then the length of the notes relative to the tempo should take care of itself via the code that's already written.

The formula we have in place is this:

```
for each instrument in *current-instruments*:
    note duration in ms = the duration of the note combined with the instrument's tempo attribute value
```

daveyarwood:
The way that this works under the hood is that the function `alda.lisp/duration` returns a function, which, when applied to a tempo, returns the duration in milliseconds.

`alda.lisp/note` is a macro that creates a note event for each instrument in `*current-instruments*`, and the duration (in ms) of the note is derived from that instrument's current tempo.

src/alda/lisp/cram.clj
crisptrutski: Whoops, yeah this seems superfluous *snip*

crisptrutski:
@daveyarwood So I think this is working now, besides the multi-instrument stuff.

Tested it on this nonsense: 

```
piano:
a b c {d e f} g
a {b c d e3}
a {b c d e}2 f
{e f g} {e {f g} {a b c {d e f}}}
```

Getting back to the duration business, think we're just missing each other slightly here - I think that without further work the instruments will *unintentionally match* durations.

Say instruments `A` and `B` are playing the same cram `{e2 f4 g g}`. `A` is currently playing 4th notes and 8th notes. When iterating through `#{A B}` in this random universe `A` iterates first.

In this implementation, a time multiplier of `1/5` is applied, because the 5 internal beats are being played in one 4th note for A. Then during playback, both instruments receive the same note calls, with an explicit duration for the first note. This means that `B` will play the phrase with the same timing as `A` instead of at twice the tempo.

crisptrutski:
Around multiple instruments however, I can't figure out how to activate more than one `*current-instruments*` via. alda code itself. Is this a "hidden feature" for programmatic usage only right now?

daveyarwood:
@crisptrutski You can do:

```
piano/guitar: c8 d e f g
```

More info here: https://github.com/alda-lang/alda/blob/master/doc/scores-and-parts.md

daveyarwood:
clicked the wrong button :)

daveyarwood:
Your `resetting` macro is cool -- we should keep it in `alda.util` maybe.

I can't help but wonder if there is a way to do this without having to do a "dry run" to get the durations and then resetting the dynamic vars. I'll think about this some more -- it's on the tip of my mind-tongue.

BTW, I pushed to the branch just now -- rebased on top of the current state of master, and also moved `alda.lisp.cram` to `alda.lisp.events.cram`. Will look at this some more in an hour or two.

crisptrutski:
Thanks for pointing out the syntax - unfortunately the "unwanted synchronisation" issue does occur. I tried a hacky solution by scaling `*time-scaling*` at play time using relative duration attributes, but that wasn't robust. Now finishing another hacky solution that sets `*time-scaling*` to a map.

Certain we can find a way that avoids the dry-run, but looks like it may require some other refactors. The use of `alter-var-root` is just quite pervasive, and I would prefer to use functions rather than macros so that crams can analyse arbitrary code without becoming its own interpreter.

Thanks for the warn on rebase, hate getting caught off guard by those :smile:

crisptrutski:
OK, supporting multiple voices at different meters now, and added a simple test song for that

crisptrutski:
The `(doseq [i *current-instruments*] (binding [*current-instruments* #{i}] ...)` pattern seems to be needed quite often - seems a bit smelly.

Perhaps attribute setters should have an arity to take the instrument, like the getters do?

daveyarwood:
Nice!

In playing that example score and playing around with this in the REPL, I think the math might be slightly off -- it sounds right for crams that last a quarter note, e.g. `{c e g}4`, but crams lasting a half note, e.g. `{c e g}2` take twice as long (i.e. they last a whole note), and crams lasting a whole note take 4 times as long. It should work out that the duration set on the cram should last as long as a simple note would.

In other words, the following two lines should be the same length:

```
piano: c2 a4 b > c1 <
piano: {c e g}2 a4 b > c1
```

crisptrutski:
Think I fixed that twice-the-time-strech-you-ordered bug, had to do with the duration reset to 1 inside crams.

To come back to that choice, feeling strongly that it was the right call now though - consider `piano/harp: {a b4 c}`. If the two instruments have a different duration when encountering the cram, then.. they could be playing `{a2 b4 c}` and `{a4 b4 c}` respectively for example. So different scale factors AND different scaled notes.

Supporting that would be even more difficult with the present approach, and the behaviour even harder for users to fathom in my estimation.

crisptrutski:
On the topic of simpler implementations - being able to turn of the implicit setting of duration when notes are played would also help open up other approaches. Maybe a `*remember-duration*` var.

daveyarwood:
Yes! This is super awesome and fun to play with. So many cool polyrhythms are at my fingertips! :computer: :notes:

> On the topic of simpler implementations - being able to turn of the implicit setting of duration when notes are played would also help open up other approaches. Maybe a *remember-duration* var.

+1 on this

daveyarwood:
(I got so excited, I forgot to say -- that totally fixed it :+1: )

crisptrutski:
@daveyarwood feel free to merge this for now, don't have the steam to look at `duration-ms` tonight

daveyarwood:
Gonna go ahead and merge this, since it works, it just needs to be performance-optimized. I'll make a separate issue for that, which I'll take a look at myself once my life stops having things going on in it (what's up with that?)

crisptrutski:
Will race you to make it fast :stuck_out_tongue:

daveyarwood:
You're on! :racehorse:

PR #118, opened 2015-10-04
Fixup earliest event calculation for no events

crisptrutski:


daveyarwood:
Thanks! I'm in the process of pushing this same commit to master, will deploy right away. :+1:

Issue #92, opened 2015-09-13
Make "final product" less(/not) reliant on Boot


daveyarwood:
I added `boot.properties` before Clojure 1.7.0 became the default Clojure version in Boot 2.2.0, as a way of enforcing that Clojure 1.7.0 be used when running Alda boot tasks, which used to be how you would run alda's `play`, `parse` and `repl` tasks, pre-`alda.cli` (i.e. instead of using the `alda` executable, you would `cd` into the Alda project directory and run `boot play --file /some/file.alda`, or `boot alda-repl`, etc.)

Since then, the latest versions of Boot (since 2.2.0) have made Clojure 1.7.0 the default `BOOT_CLOJURE_VERSION`, and the usage instructions for Alda has moved from cloning the repo and running Boot tasks to copying a provided executable Boot script (`bin/alda`) to your `$PATH` and using the `alda` command as an entrypoint into the tasks defined in the `alda.cli` namespace.

Bearing this in mind, does it make sense for us to still have a `boot.properties` file? With Boot steadily evolving and improving, I'm not sure I like the idea of pinning the Alda project to a specific version of Boot, when really all we want to enforce is Clojure 1.7.0 as a dependency.

crisptrutski:
In my opinion repeatability is worth it.

Running `boot -V > boot.properties` every so often isn't too onerous, and if you're worried about forgetting and being left behind you could add a CI test comparing asserting that `boot.properties` is up-to-date.

daveyarwood:
I guess my question is more about whether we still *need* boot.properties, since we have moved away from using Boot tasks as the way to run Alda in the wild. For the "end product" version of Alda (i.e. the executable script), it shouldn't matter what version of Boot you have or what you might have `BOOT_CLOJURE_VERSION` set to -- only that the JVM (I guess the class loader?) in which Alda is running has Clojure 1.7.0 available to it.

crisptrutski:
Perhaps it's worth exploring a `bin/alda` that doesn't even require boot on the client (it just pulls the JAR and runs it). Flipside, you since `2.0.0-rc13` boot has supported launching pods with specific version of Clojure (so move `alda` task into one)

From the perspective of developers collaborating, and being able to cut long term release branches one day, I think the properties file should stay in the repo, even if it plays no role in the distributed artifact.

daveyarwood:
OK -- I'm convinced that boot.properties can stay -- it makes sense from a development perspective :)

[...]

PR #121, opened 2015-10-06
Package alda as an uberjar

jeluard:
Removed dependencies on `boot` by porting `alda.cli` to `tools.cli` and investigate benefits of using uberjar to improve `alda` startup time.

Overall this improves the startup time but it's still slow on my machine. Doing some more tests it looks like using a regular jar with dependencies loaded via `java -classpath` provides better performance. I will update on this.

This relates to #92.


crisptrutski:
Looking good :sunglasses:

daveyarwood:
This is awesome. Thanks @jeluard !

PR #172, opened 2016-01-13
Build uberjar with JDK 1.7 support

### LINKED ISSUE #171: Support older Java versions
Alda isn't working for me with OpenJDK.

OS: Debian 8

Java (`$ java -version`):

```
java version "1.7.0_91"
OpenJDK Runtime Environment (IcedTea 2.6.3) (7u91-2.6.3-1~deb8u1)
OpenJDK 64-Bit Server VM (build 24.91-b01, mixed mode)
```

I get this error:

```
Exception in thread "main" java.lang.UnsupportedClassVersionError: alda/Client : Unsupported major.minor version 52.0
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
	at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
	at java.security.AccessController.doPrivileged(Native Method)
	at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:482)
```

I have Oracle's JVM installed outside my PATH for playing Minecraft, so I can just set up an alias (`alias alda='/usr/local/java/bin/java -jar ~/.local/bin/alda $@`). However, it'd be nice to use the open-source version. Anyone know what's going on?

daveyarwood:
Closes #171 

`boot build` will now build uberjars (and therefore binaries) that are compatible with JDK 1.7+, making it so that people with older Java versions can use Alda.

I tried going back to JDK 1.6, but there are a couple incompatible things in the Java client source that would need to be resolved, and I'm not sure if it's worth the effort at this point, with JDK 1.6 on its way out. We can always revisit in the future if there is a need for Java 6 compatibility.

Blockers for Java 6 compatibility:

1) The switch statement in Client.java operates on strings, which is not supported in Java 6. We could just make it a series of if/else if statements.

2) The use of `.inheritIO()` in Util.java here:

     public static void runProgramInFg(String... args)
     throws IOException, InterruptedException {
       new ProcessBuilder(args).inheritIO().start().waitFor();
     }

   I'm sure there is some pre-JDK7 workaround for this.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -54,6 +54,10 @@
       version)))
 
 (task-options!
+  javac   {:options ["-source" "1.7"
+                     "-target" "1.7"
+                     "-bootclasspath" (System/getenv "JDK7_BOOTCLASSPATH")]}
+
   pom     {:project 'alda
            :version alda.version/-version-
            :description "A music programming language for musicians"
@@ -103,6 +107,21 @@
                           alda.parser.examples-test
                           }})
 
+(deftask assert-jdk7-bootclasspath
+  "Ensures that the JDK7_BOOTCLASSPATH environment variable is set, as required
+   to build the uberjar with JDK7 support."
+  []
+  (with-pre-wrap fileset
+    (assert (not (empty? (System/getenv "JDK7_BOOTCLASSPATH")))
+            (str "Alda requires JDK7 in order to build its uberjar, in order "
+                 "to provide out-of-the-box support for users who may have "
+                 "older versions of Java. Please install JDK7 and set the "
+                 "environment variable JDK7_BOOTCLASSPATH to the path to your "
+                 "JDK7 classpath jar, e.g. (OS X example) "
+                 "/Library/Java/JavaVirtualMachines/jdk1.7.0_71.jdk/Contents/"
+                 "Home/jre/lib/rt.jar"))
+    fileset))
+
 (deftask dev
   "Runs the Alda server for development.
 
@@ -133,7 +152,11 @@
 (deftask package
   "Builds an uberjar."
   []
-  (comp (javac) (pom) (uber) (jar)))
+  (comp (assert-jdk7-bootclasspath)
+        (javac)
+        (pom)
+        (uber)
+        (jar)))
 
 (deftask build
   "Builds an uberjar and executable binaries for Unix/Linux and Windows."


crisptrutski:
Don't know much about this, but curious if you tried cross compiling (just setting target, and using JDK 1.8)

daveyarwood:
Interesting... I don't know much about that either. I'll give it a whirl when I have a minute. Would that save us the hassle of specifying the bootclasspath?

daveyarwood:
Hmm... seems like it's not possible to compile with `-target 1.7` if your system is 1.8.

I tried just now and got this warning:

```
javacTask: target release 1.7 conflicts with default source release 1.8
```

Then when I ran the resulting Alda executable, I got:

```
Error: Could not find or load main class alda.Client
```

Changing my JVM to 1.7 (via jenv) and doing the same thing, it compiles and runs just fine.

Interestingly, tried `-target 1.7 -source 1.7` and there was a warning `WARNING: bootstrap class path not set in conjunction with -source 1.7`, but it compiled nonetheless and ran successfully on both 1.7 and 1.8! I guess maybe it uses the 1.8 bootclasspath, which works when compiling 1.7 source?

crisptrutski:
Thanks for exploring! Interesting to understand this cross compilation. I think you're summary is on the money - it's using the 1.8 bootstrap classes (*wave hands about rt.jar here*), and it turns out this is fine in most cases but can lead to very sad situations when your program gets large enough to pull in corner classes that have changed.

Looked around for a concrete example and found this interesting article: http://www.draconianoverlord.com/2014/04/01/jdk-compatibility.html

daveyarwood:
Ah -- yeah, that makes sense. Probably best to compile using the target version's bootclasspath, so the compiler will catch any 1.8-only things that might sneak into the source code.

PR #185, opened 2016-02-14
Add an autoupdating script

### LINKED ISSUE #82: Improve updating mechanism
For some reason, `alda` isn't updating to latest version.  I went in and changed `LATEST` to `0.5.4` and it downloads the new version, but then if I change it back to `LATEST` it goes back to `0.5.2`.  Is there a way to force boot to search for a newer update?

Also would there be a performance gain if alda was installed locally (behind a flag perhaps)? Then, when running `alda` you could either have it check for a local install or have a flag that skips the update check.

jgkamat:
Still a WIP, but the basic functions are working as of now as far as I know. This adds an `alda update` command that updates alda based off of the latest release. I've only tested this on linux, but I think this should work on windows/mac too (but it needs to be tested).

Currently, it simply overwrites the current alda executable with the one from the latest release (and does no checking to see if an update is needed or not). I still need to figure out how to do that (maybe starting the server to find the version). Another problem is that the version tags don't match up with the version reported by alda version (so getting the version from the server might not help too much either way).
    
As far as I know, I didn't need to make the downloaded file executable (It seemed to preserve the permissions of the old alda file? I'm going to double check this later).

Overview:
- Query GH Api to get latest release tag/download url
- Remove .exe if it exists from download url (for unix compatibility)
- Add .exe back to download URL if on windows
- Download release to path specified by current alda executbale
    
Let me know what feedback you have, and I'll get to fixing it!

Also, one more thing, something about installing launch4j should probably be added to the [developers guide](https://github.com/alda-lang/alda/blob/master/doc/development-guide.md#testing-changes).

Closes #82


daveyarwood:
Wow -- awesome first PR! Thanks!

I will take a close look at this later when I have some time. I'm on a Mac, so I'll test it out and let you know how it works for me.

> Currently, it simply overwrites the current alda executable with the one from the latest release (and does no checking to see if an update is needed or not). I still need to figure out how to do that (maybe starting the server to find the version).

I was thinking we could move the version number to a dedicated file like `version.txt` on the resource path, and that way both the Clojure server and Java client can read from it. An alternative would be for the Java client to require `alda.version/-version-` directly, but then it would take at least a few seconds for the Clojure runtime to start every time the client wants to check the version. Putting it in a file on the resource path is the best option, I think.

It's probably best that we not read the version number from the server anyway, as the server can potentially be a different version from the client. The way it works is, when the client starts up a server, it spawns a new process of itself. So it's possible to do something like this:

* Have Alda 1.0.0-rc6 installed.
* Run `alda up` -- this starts an Alda server running 1.0.0-rc6 (same as client) on port 27713.
* Leave the server running.
* Download and install Alda 1.0.0-rc7.
* Run `alda version` -- this asks the server running on the default port (27713) for its version -- it will report 1.0.0-rc6 despite the client being 1.0.0-rc7.

Now that I think of it, maybe we should have the client report both its own version and the server version when you run `alda version`. This would make it easier to catch when your server and client are not running the same version of Alda. (We may also want to print a reminder after updating, something like `alda has been updated to 1.0.0-rc7. If you have any currently running servers, you may want to restart them so that they are running the latest version.` Better yet, we could check to see if any servers are running and only display this reminder if there are.)

> As far as I know, I didn't need to make the downloaded file executable (It seemed to preserve the permissions of the old alda file? I'm going to double check this later).

I think when I upload the executables (which are already marked executable), GitHub preserves the `+x` permission, but it would probably be good if we set the file to executable anyway, just in case.

> Also, one more thing, something about installing launch4j should probably be added to the developers guide.

That's a good call -- I'll take care of that.

Will give more feedback when I have some time to try it out. Thanks again!

client/src/alda/Util.java
```diff
@@ -137,6 +146,114 @@ public static String getProgramPath() throws URISyntaxException {
...
+      return;
+    }
+
+    // Turn api result into version numbers and links
+
+    Gson gson = new Gson();
+    JsonObject job = gson.fromJson(apiResult, JsonObject.class);
+
+    // Gets the download URL. This may have ...alda or ...alda.exe
+    String downloadURL = job.getAsJsonArray("assets").get(1).getAsJsonObject().get("browser_download_url").toString();
```
daveyarwood: It looks like the `get(1)` part of this will always grab the second asset, so we are relying on that being `alda.exe`. I wonder if maybe we should check whether the OS is Windows, then iterate through the assets until we find the download URL that ends with either "alda" or "alda.exe"?

daveyarwood:
Aside from the note above, this looks good!

It did successfully update for me on OS X. :+1:  Interestingly, it looks like because the file it's replacing is already executable, the new file is becomes executable automatically. However, if I run `alda update`, then delete the current version before confirming the update, it will download the file, but the file it downloads is not executable. So, I'll add as a TODO item that we should make the new file executable after downloading it.

TODO before merging:

- [ ] Explicitly look for an asset ending in either "alda" or "alda.exe" (see note above)
- [ ] Make the downloaded file executable (even if it already is, it doesn't hurt to make sure)
- [ ] After downloading, print something like `Updated to alda 1.0.0-rc7.` so the user knows the update was successful.
- [ ] Check to see if the version the user has is already up-to-date. If so, print something like `alda is up-to-date.` and exit.

jgkamat:
Sure thing! I'm going to make a loop to cycle through the assets and search for exe if on windows (and no exe if on unix) to solve the above problem. That's much cleaner than what I was doing earlier.  Do you know which folder the resource path was located? I'm having a bit of trouble trying to find it (to replace the clojure version file with a common version.txt).

I'll probably be working on these issues tomorrow.

daveyarwood:
The resource paths are set here in build.boot: https://github.com/alda-lang/alda/blob/master/build.boot#L3-L4

I would add a top-level `resources` directory (which we can use for any shared client/server resources) and move the version to `resources/version.txt`.

Then we can replace the Clojure code in [alda.version](https://github.com/alda-lang/alda/blob/master/server/src/alda/version.clj) with this:

```
(ns alda.version
  (:require [clojure.java.io :as io]
            [clojure.string  :as str])

(def ^:const -version-
  (str/trim (slurp (io/resource "version.txt"))))
```

jgkamat:
I've just pushed commits that fix the problems that we had! It now checks to see if the version string matches the tag found online (if not, trigger an update). It also loops through the entries to find the correct download instead of the string manipulation I did before. The messages might need a bit more polish, but the core features are there.

I also added a client/server printout to `alda version`. The update command dosen't make use of this though, but it does print out the warning if the update was completed.

It would be nice to test this out on windows so, here are compiled files for windows/unix that have their version hacked to be 1.0.0-rc6, to test updating.
[alda_update.zip](https://github.com/alda-lang/alda/files/130060/alda_update.zip)

daveyarwood:
This looks great :+1: Will merge soon.

@pkobrien: Any interest in testing this update functionality on Windows? If you download the patched version of Alda from alda_update.zip above, you should be able to do the following:

* Run `alda.exe up` to start the server.
* Run `alda.exe version` -- it should say 1.0.0-rc6.
* Run `alda.exe down` to stop the server.
* Run `alda.exe update` -- it should fetch the latest Alda release (1.0.0-rc7) and replace alda.exe with the alda.exe from that release.
* Run `alda.exe start`, then `alda.exe version` again -- it should say 1.0.0-rc7.

daveyarwood:
I'm going to go ahead and merge this -- it seems like it should do the right thing™ for Windows, and we can definitely revisit if any Windows-related updating issues surface.

@jgkamat thanks again!

PR #201, opened 2016-03-09
Add a slew of validations.

### LINKED ISSUE #195: Attributes do not check for valid input data
I was playing around with attributes recently, and I found something interesting in the way attributes handle input data. I noticed that almost all attributes do not check for invalid input data, only for an invalid data type (they also always throw java/clojure exceptions as well; never alda ones). This leads to some interesting cases:

1. Duplicate/Invalid key signatures  
Key signatures that are invalid have no effect. `piano: (key-sig "x+") o4 d e f g a b > c d` runs exactly the same as `piano: o4 d e f g a b > c d`. Key signatures that are duplicates choose the latter note. `piano: (key-sig "f+ f- g+ g-") o4 d e f g a b > c d` runs exactly the same as `piano: (key-sig "f- g-") o4 d e f g a b > c d`
2. Invalid durations  
Durations should be positive. Nonpositive durations have no effect. `piano: (duration 0) o4 d e f g a b > c d` runs exactly like `piano: o4 d e f g a b > c d`
3. Invalid Octaves  
Octaves should be integers. `piano: (octave 4.5) d e f g a b > c d` runs the code as if it was running `piano: (octave 4) d++++++ e++++++ f++++++ g++++++ a++++++ b++++++ > c++++++ d++++++`. This *kind of* makes sense due to how octaves work philosophically, but it's inconsistent with the syntax (`piano: (octave 4.5) d e f g a b o5.5 c d` throws an Alda syntax error)
4. Invalid Tempos  
Tempos should be positive real numbers. running `piano: (tempo 0) o4 d e f g a b > c d` throws a divide by zero error, however negative tempos do not throw errors. I haven't figured out what negative tempos do; they behave very weirdly.
5. Invalid Quantization  
Quantization should be between 0 and 100. Negative quantizations throw an error, however `piano: (quant 1000) d e f g a b o5 c d` runs exactly like `piano: (quant 100) d e f g a b o5 c d` and doesn't throw an error. 

Panning, volume, and track-volume check for valid input data properly.

jgkamat:
Closes #195

Adds a bunch of validations for a couple functions. I think I've got the desired behavior except for `(duration #)`. For some reason, the assertion only works on `(set-duration #)`, and everything I tried to rename it to duration (and alias to set-duration) dosen't seem to work. I'm hoping you'll be able to figure out what I'm doing wrong there. Seems like things like `a-1` still work, but `a0` is caught. 

Let me know if I'm doing anything wrong, this is my first time writing clojure :smiley:

### DIFF
--- a/server/src/alda/lisp/attributes.clj
+++ b/server/src/alda/lisp/attributes.clj
@@ -14,15 +14,28 @@
   {:pre [(<= 0 x)]}
   (constantly (/ x 100.0)))
 
+;; Validation that the input is an integer value
+(defn- pos-integer[x]
+  {:pre [(and (integer? x)
+              (pos? x))]}
+  (constantly x))
+
 (defattribute tempo
   "Current tempo. Used to calculate the duration of notes."
-  :initial-val 120)
+  :initial-val 120
+  :transform pos-integer)
 
 (defattribute duration
   "Default note duration in beats."
   :initial-val 1
   :fn-name set-duration
+  ;; :aliases [:duration]
   :transform (fn [val]
+               { :pre [(or
+                        (map? val)
+                        (and (number? val)
+                             (pos? val)))]}
+
                (constantly (if (map? val)
                              (:value val)
                              val))))
@@ -31,7 +44,7 @@
   "Current octave. Used to calculate the pitch of notes."
   :initial-val 4
   :transform (fn [val]
-               {:pre [(or (number? val)
+               {:pre [(or (integer? val)
                           (contains? #{:down :up} val))]}
                (case val
                 :down dec
@@ -69,6 +82,22 @@
   :initial-val 0.5
   :transform percentage)
 
+(defn- validate-str-key-sig
+  "Validates the current key-sig. Checks for:
+
+  1. No duplicate letters, ie: a- a+
+  2. No letters out of range a-g
+
+  If all tests pass, return true"
+  [key-sig]
+  ;; Get a version of key-sig with only characters
+  (let [clean-str (apply str (filter #(Character/isLetter %) key-sig))]
+    (and
+     ;; Check to see if any chars are not a-g
+     (empty? (re-find #"[^a-gA-G]" clean-str))
+     ;; No duplicates
+     (= (count (distinct clean-str)) (count clean-str)))))
+
 (defn- parse-key-signature
   "Transforms a key signature into a letter->accidentals map.
 
@@ -79,6 +108,9 @@
    If the key signature is provided as a string, e.g. 'f+ c+ g+', then it is
    converted to a letter->accidentals map."
   [key-sig]
+  {:pre [(or (not (string? key-sig))
+             (validate-str-key-sig key-sig))]}
+
   (constantly
     (cond
       (map? key-sig)
@@ -87,7 +119,7 @@
       (string? key-sig)
       (into {}
         (map (fn [[_ _ letter accidentals]]
-               [(keyword letter) 
+               [(keyword letter)
                 (map {\- :flat \+ :sharp \= :natural} accidentals)])
              (re-seq #"(([a-g])([+-=]*))" key-sig)))
 
--- a/server/src/alda/lisp/attributes.clj
+++ b/server/src/alda/lisp/attributes.clj
@@ -33,7 +33,7 @@
   :transform (fn [val]
                {:pre [(or
                        (map? val)
-                       (and (number? val) (not (neg? val))))]}
+                       (and (number? val) (pos? val)))]}
 
                (constantly (if (map? val)
                              (:value val)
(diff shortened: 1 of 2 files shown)


daveyarwood:
This looks great! I only see a few issues, and they are mostly just style points. I will leave comments on individual lines.

The duration issue is a result of this hack that we can remove now: https://github.com/alda-lang/alda/blob/master/server/src/alda/lisp/model/duration.clj#L73  `duration` calls `set-duration` under the hood, but only if the duration is positive. Since we are doing all the validation in `set-duration` now, we can change this line to just `(set-duration beats)`.

server/src/alda/lisp/attributes.clj
```diff
@@ -14,15 +14,28 @@
   {:pre [(<= 0 x)]}
   (constantly (/ x 100.0)))
 
+;; Validation that the input is an integer value
+(defn- pos-integer[x]
```
daveyarwood: There should be a space between the function name and the vector of arguments: `pos-integer [x]`

server/src/alda/lisp/attributes.clj
```diff
@@ -14,15 +14,28 @@
...
   "Default note duration in beats."
   :initial-val 1
   :fn-name set-duration
+  ;; :aliases [:duration]
   :transform (fn [val]
+               { :pre [(or
+                        (map? val)
+                        (and (number? val)
+                             (pos? val)))]}
+
```
daveyarwood: 1. In a Clojure map, there should usually be no space between the opening curly brace and the first thing in it.

2. Re: line breaks / spacing, this is a matter of preference, but I would write it like this:

```clojure
(fn [val]
  {:pre [(or (map? val)
             (and (number? val) (pos? val)))]}
```

server/src/alda/lisp/attributes.clj
```diff
@@ -69,6 +82,22 @@
...
+  1. No duplicate letters, ie: a- a+
+  2. No letters out of range a-g
+
+  If all tests pass, return true"
+  [key-sig]
+  ;; Get a version of key-sig with only characters
+  (let [clean-str (apply str (filter #(Character/isLetter %) key-sig))]
+    (and
+     ;; Check to see if any chars are not a-g
+     (empty? (re-find #"[^a-gA-G]" clean-str))
```
daveyarwood: 1. We should actually only allow `a-g`. Something like `"F+ C+ G+"` should throw an error.

2. `re-find` will return `nil` if it doesn't find any instances of the regex in the string. So, it would be more idiomatic to use `not` here, instead of `empty?` which has slightly different semantics.

3. This is also a matter of personal preference, but I would include the first condition on the same line of the `and`, like this:

    ```clojure
    (and (not (re-find #"[^a-g]" clean-str))
         (= (count (distinct clean-str)) (count clean-str)))
    ```

elyisgreat:
Can someone summarize the new validations? I don't speak clojure :(

jgkamat:
Here's whats changed (still working on the finishing touches):
Errors on:
```
(key-sig "a+ a+")
(key-sig "z+")
(duration -1)
(duration 0)
(octave 1.1)
(tempo 0)
(tempo -1)
(tempo 1.1)
```
While keeping the rest of the behavior intact

elyisgreat:
Tempo 1.1 is not a problem.

daveyarwood:
@elyisgreat You're right, arguably we should allow non-integers for tempo. The key thing is just that tempo is a positive number.

jgkamat:
I'll fix that in my next commit, thanks!

jgkamat:
I think I addressed all the problems, but the check for `(duration 0)` seemed to be causing problems (so I removed it), when I ran `(duration (ms 100))` that check would cause it to fail. If you can tell me what I'm doing wrong, I'll fix it! :smile: 

The tests seem to be passing for me now.

server/src/alda/lisp/model/duration.clj
```diff
@@ -70,7 +70,7 @@
         ms    (apply + (for [{:keys [type value]} note-lengths
                              :when (= type :milliseconds)]
                          value))]
-    (when (pos? beats) (set-duration beats))
+    (set-duration beats)
```
daveyarwood: I didn't think about millisecond durations. This is something we should probably address in the future, but for now, we can continue to ignore them.

I think changing this line to `(when beats (set-duration beats))` will do the trick.

jgkamat:
Sorry for the spotty turnarounds, I've been kind of busy today. I haven't gotten a chance to test the latest commit at all (I can't seem to get the build working on my second laptop, i'll debug that later on). I'll test this out when I get back, and comment what I find out.

jgkamat:
The `(duration (ms 100))` test seems to be failing, but otherwise, it looks good.

daveyarwood:
@jgkamat I just played around with it, and I've got a fix ready. I'm going to merge your PR, and then apply my change in a new commit.

The issue was that beats is never actually `nil` -- it defaults to 0 if there are no note-lengths expressed in beats (for example if you do `(duration (ms 1000))`). So if we do `(when beats (set-duration beats))`, then it will try to set the duration to 0 beats, triggering the assertion error because the number isn't positive.

Thanks for this PR, nice work! :+1: :smile: :+1:

Issue #174, opened 2016-01-18
Client: add command-specific help text


daveyarwood:
You can get an overview of all of the available commands and options by typing `alda -h` or `alda --help`.

It would be nice if you could get an overview of available options for each command by typing, e.g. `alda parse -h`.

(There might be some convenient built-in way to do this with JCommander.)

jgerman:
I got bored and implemented this here: https://github.com/alda-lang/alda/pull/215

PR #215, opened 2016-04-01
Adds Command Specific Help

jgerman:
This adds command specific help. Due to JCommander shadowing the original command class there wasn't an entirely acceptable solution, so I took the straightforward route. Commands inherit help behavior from a base class and we check for help in the big switch statement in main(). 

It's not quite as succinct as I'd like but it works.

### DIFF
--- a/client/src/alda/Client.java
+++ b/client/src/alda/Client.java
@@ -1,5 +1,7 @@
 package alda;
 
+import java.util.Map;
+import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


daveyarwood:
This looks pretty straightforward to me. Not bad! :smile: :+1: 

Thanks for the PR -- this will make it a lot easier to explore all of the different options available when using the Alda CLI client.

Closes #174.

PR #218, opened 2016-04-14
Updated the Windows installation instructions

### LINKED ISSUE #217: why move alda.exe to the windows system root?
This sounds like an incredibly bad idea, instead it would serve people better to explain how to put alda.exe in its own `C:\program files\alda` or similar folder, because `program files` (or `program files (x86)` for 32 bit apps) is where windows programs are supposed to go.

And then if you absolutely need to run it from anywhere (which most windows users won't, because they have no reason to use the command line for anything), explain how to update their PATH, which (again) is the proper way to do things.

And the people that *do* use the command line will be happy to just add another dir to the PATH for the convenience, so trying to get people to copy your executable into the system dir, no matter how you look at it, is not good advice, and worth doing right instead.

Pomax:
fixes #217 

I've updated the README.md with instructions on how to manually "install" the `alda.exe` program following standard Windows conventions, although I am assuming it's a 32 bit, rather than a 64 bit executable. If it's 64 bit then it'll have to go into `Program Files`, not `Program Files (x86)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,9 +110,15 @@ For more examples, see these [example scores](https://github.com/alda-lang/alda/
 
 * Go to the [latest release](https://github.com/alda-lang/alda/releases/latest) page and download `alda.exe`.
 
-* Make the `alda` command available by moving `alda.exe` to your system root path:
+* Make the file executable:
+  * Go to your downloads folder, right click `alda.exe` to open up its file properties, and click `unblock`
 
-        C:\> move alda.exe %SystemRoot%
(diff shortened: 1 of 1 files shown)


daveyarwood:
This looks great -- thanks!

I tried the method described in [this StackOverflow answer](http://superuser.com/a/889267/485305), and it looks like `alda.exe` is a 64-bit executable. Would you mind updating your instructions to use `Program Files`?

daveyarwood:
Actually, scratch that -- it's x86!

PR #221, opened 2016-04-19
alda+.exe

lifeluvr:
A self-contained, stand-alone, nothing-to-install, run it and go app for Windows.
Delete it and it's gone!
This is a prototype and I'm looking for suggestions to improve it.
Some easy options are:
A different icon
Custom welcome message just above the 'alda repl' initial display.
Can also run the above command 'hidden' so the user cant see it.
Keep the tmp file or create the file in current directory (right now it's set to delete the file when closed).
I can also add other files for the program to use or enhance alda.


daveyarwood:
Thanks for the PR -- I appreciate your contribution, and this is definitely a cool idea, but I will have to decline for the two reasons I mentioned earlier:

1. I would like to keep the official install process more in line with the philosophy of the command-line. We already release an `alda.exe` which Windows users can install without too much difficulty, and then use freely, in a modular way, and be able to use all of the available commands that the `alda` client offers.

2. It would be a hassle to have to generate an `alda+.exe` as well as an `alda.exe` with every release.

Also, an important note -- please wait for approval before merging pull requests, or better yet, allow me to merge them if I approve :)

lifeluvr:
ooops my bad...
After your last email I thought that's what you wanted..
Sorry about that ;)

On 4/18/16, Dave Yarwood <notifications@github.com> wrote:
> Thanks for the PR -- I appreciate your contribution, and this is definitely
> a cool idea, but I will have to decline for the two reasons I mentioned
> earlier:
>
> 1. I would like to keep the official install process more in line with the
> philosophy of the command-line. We already release an `alda.exe` which
> Windows users can install without too much difficulty, and then use freely,
> in a modular way, and be able to use all of the available commands that the
> `alda` client offers.
>
> 2. It would be a hassle to have to generate an `alda+.exe` as well as an
> `alda.exe` with every release.
>
> Also, an important note -- please wait for approval before merging pull
> requests, or better yet, allow me to merge them if I approve :)
>
> ---
> You are receiving this because you modified the open/close state.
> Reply to this email directly or view it on GitHub:
> https://github.com/alda-lang/alda/pull/221#issuecomment-211668682


-- 
   ***Eric Payne***
   1-619-786-4687
erdapa1@gmail.com
   www.lifeluvr.org

daveyarwood:
No, I meant you should make it a repo under your own username. Sorry for
the confusion!

Issue #232, opened 2016-05-28
FluidR3 installation script fails if the Gervill directory doesn't exist


another-dave:
### Steps to reproduce:

Following the installation instructions in the README, after downloading Alda, run the `$ALDA_HOME/scripts/install-fluidr3` script without a `.gervill` folder present in your home directory.

### Outcome:

Script fails with error:

> `cp: /Users/[your-user]/.gervill/soundbank-emg.sf2: No such file or directory`

### Solution

Update script to make `.gervill` directory if it doesn't exist

*(Tested on OSX 10.9; Java SE7; Bash 3.2)*

daveyarwood:
Good catch! I wonder why the `.gervill` directory was missing in your case... maybe Java only sets that up the first time you run a Java program that uses the built-in MIDI synth?

Would you mind confirming that your version of Java uses `~/.gervill`? You can tell if it does if, after putting the FluidR3 soundfont in that directory (i.e. run the script) and using Alda, the MIDI instruments sound like the FluidR3 ones. To hear the difference between FluidR3 and the Java default soundfont, you can get the default soundfont back by running `rm ~/.gervill/soundbank-emg.sf2` -- Java should replace that file with the default soundfont the next time you use it.

another-dave:
Yes, just tried removing it there & it falls back to the default soundfont (wow, that difference is massive!). Also, with it removed & trying Alda again, the JVM does create the directory on the fly after that.

So guess it's only an issue for people in my shoes where hadn't done any MIDI work before in Java & trusted that the default soundfont would be rubbish & upgraded first go :)

daveyarwood:
I see, interesting. Thanks for the PR, that should be helpful for other people in your boat!

another-dave:
No worries, cheers Dave! Hopefully will commit some PRs against the main code base in the future :) Came across the project on the [Brave Clojure](http://open-source.braveclojure.com/) website, looks really interesting!

daveyarwood:
That's awesome, thanks! Let me know if you have any questions about how this stuff works, I'd be happy to help.

PR #233, opened 2016-05-28
#232 Updating installation script to resolve missing Gervill directory

another-dave:
Providing a PR to resolve [issue 232](https://github.com/alda-lang/alda/issues/232) — that the FluidR3 script will fail if the `~/gervill`directory is missing.

(Please let me know if you need any amends)


daveyarwood:
Looks good 👍

Issue #208, opened 2016-03-13
Improving song loading time


elyisgreat:
I know you've [talked about the starting time before](https://github.com/alda-lang/alda/issues/74), however I think that it pertained more to when using commands, not in the REPL. I've noticed that for longer scores (>2 minutes), it can take several seconds to load it in, and a few more between when the `:play` command is run and when the song actually plays (This happens with `alda play` as well). I think improving the loading time would greatly improve writing productivity. Of course, this is no easy task, as this could be a huge parser issue. So here  I leave the issue open.

daveyarwood:
Parsing is the bottleneck. I was talking to one of my coworkers, and he pointed out that the issue could be ambiguity in the parser. I think that's probably the case. Going through and trying to eliminate any ambiguity in the parser could help a lot.

daveyarwood:
@elyisgreat Whenever you have a minute, give alda 1.0.0-rc19 a try, and let me know your thoughts. This release features some significant parser improvements, making loading time faster. There is still more we could do, but I think we are off to a good start for now.

daveyarwood:
Moved to https://github.com/alda-lang/alda-core/issues/18.

Issue #131, opened 2015-10-15
Variables


daveyarwood:
Original idea (originally proposed by me in #7, now being broken out into its own issue):

> **Variables**

> Variables should work like they do in Lilypond. Any piece of code can be enclosed in brackets and defined as a variable, then called in the code any number of times, at any point in time.

> Desired syntax:

> ```
# Defining a variable: 
riffA = [f8 f g+ a > c c d c <]
riffB = [b-8 b- > c+ d f f g f <]
riffC = [> c8 c d+ e g g a g < ]
riffD = [f8 f g+ a > c c d < b > | c c < b- b- a a g g]

> # calling a variable
$riffA # expands to "f8 f g+ a > c c d c <"

> # using variables + repeats
($riffA)x4 
($riffB)x2  ($riffA)x2
$riffC $riffB $riffD

> # bonus: conglomerate variables into other variables!
rockinRiff = 
  [ ($riffA)x4 
    ($riffB)x2  ($riffA)x2
    $riffC $riffB $riffD    ] 

> guitar/saxophone: 
    ($rockinRiff) x9999999
```

That was from January, and since then Alda's syntax has evolved a little. Parentheses are no longer available in the way proposed above (parentheses now mean inline Clojure code).

New proposed syntax, incorporating ideas proposed by @MillerMark in #7:

```
# Defining a variable: 
riffA = f8 f g+ a > c c d c <
riffB = b-8 b- > c+ d f f g f <
riffC = > c8 c d+ e g g a g <
riffD = f8 f g+ a > c c d < b > | c c < b- b- a a g g

# calling a variable
riffA # expands to "f8 f g+ a > c c d c <"

# using variables + repeats
riffA*4
riffB*2 riffA*2   # or riffB** riffA**
riffC riffB riffD

# bonus: conglomerate variables into other variables!
rockinRiff = [ 
    riffA*4
    riffB*2 riffA*2
    riffC riffB riffD
] 

guitar/saxophone: 
    rockinRiff*9999999
```

Key points:
- `$` is not needed before variable names -- whether a variable is being defined or evaluated can be left to context
- brackets around variable value can be omitted when it's all on one line -- cleaner than always requiring them
- `*2` syntax instead of `x2` -- easier to differentiate visually, and makes it possible to omit the space between "thing-being-repeated" and "times-N"

I'm totally sold on this new syntax. It cleaner-looking, easier to read, and quicker to type. A little added complexity in the parser is well worth the simpler syntax that it gives the user.

[...]

PR #238, opened 2016-06-03
Parser optimizations

daveyarwood:
I made a pass through each of Alda's parsers with `instaparse.core/parses` to determine whether there was any ambiguity contributing to slow parse times noted in #208. I did find a few places where there was ambiguity in the grammar -- see the commit messages for more details.

Unfortunately, this is only a minor improvement over what we have now. The major example of a slow-parsing score we have now is the Bach cello suite example. Currently on my computer, that score takes about 800-900 ms to parse. With the improvements on this branch, it takes about 700-800 ms which is barely a difference.

I think the next step is going to be considering some of the other suggestions in the [instaparse wiki's performance tips](https://github.com/Engelberg/instaparse/blob/master/docs/Performance.md#performance-tips). This PR should make performance tip 5 no longer an issue, but maybe performance tip 6 is applicable.

One thing I started playing around with was commenting out certain parts of the parsing pipeline and using `clojure.core/time` to profile the different stages. On my computer, it took about 300 ms just to remove the comments and parse Clojure expressions, which seems fishy.

(WIP PR -- will leave open, at least for now. I'd love some feedback, if anyone reading this happens to be knowledgeable about parsing and/or can help determine the causes of Alda's slowness in parsing larger scores.)

### DIFF
--- a/server/grammar/duration.bnf
+++ b/server/grammar/duration.bnf
@@ -1,6 +1,5 @@
 duration                = (note-length | seconds | milliseconds)
                           (<ows> barline? <ows> subduration)*
-                          (barline? <ows> slur)?
 <subduration>           = tie <ows> barline? <ows>
                           (note-length | seconds | milliseconds) <ows>
 
--- a/server/grammar/duration.bnf
+++ b/server/grammar/duration.bnf
@@ -1,8 +1,8 @@
 duration                = (note-length | seconds | milliseconds)
-                          <ows> barline? <ows> subduration* slur?
+                          (<ows> barline? <ows> subduration)*
+                          (barline? <ows> slur)?
 <subduration>           = tie <ows> barline? <ows>
                           (note-length | seconds | milliseconds) <ows>
-                          barline? <ows>
 
 seconds                 = positive-number <"s">
 milliseconds            = positive-number <"ms">
--- a/build.boot
+++ b/build.boot
@@ -21,7 +21,6 @@
                   [io.aviso/pretty        "0.1.20"]
                   [jline                  "2.12.1"]
                   [org.clojars.sidec/jsyn "16.7.3"]
-                  [org.clojure/clojure    "1.7.0"]
                   [potemkin               "0.4.1"]
                   [ring                   "1.4.0"]
                   [ring/ring-defaults     "0.1.5"]
--- a/server/grammar/comments.bnf
+++ b/server/grammar/comments.bnf
@@ -1,4 +1,4 @@
 score         = (non-comment | <comment> | clj-expr)*
 comment       = <"#"> #".*" <#"(\n|\r|$)+">
-<non-comment> = !("#" | "(" | ")") #".|\n|\r"
+<non-comment> = #"[^\#\(\)]+"
 
--- a/server/grammar/score.bnf
+++ b/server/grammar/score.bnf
@@ -5,5 +5,5 @@ calls             = name
                     (<ows> <"/"> <ows> name)*
                     (<ows> nickname)?
                     <ows> <":">
-<music-data>      = !part #".|\n|\r"
+<music-data>      = !calls #"[a-zA-Z]*[^a-zA-Z]*([a-zA-Z][^a-zA-Z]+)*"
 
--- a/server/grammar/score.bnf
+++ b/server/grammar/score.bnf
@@ -1,4 +1,4 @@
-score             = <ows> header? <ows> part* <ows>
+score             = <ows> header? part*
 header            = !calls <ows> music-data+
 part              = calls <ows> music-data*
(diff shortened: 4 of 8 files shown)


server/grammar/duration.bnf
```diff
@@ -1,8 +1,7 @@
 duration                = (note-length | seconds | milliseconds)
-                          <ows> barline? <ows> subduration* slur?
+                          (<ows> barline? <ows> subduration)*
```
aengelberg: I believe this line is ambiguous; whitespace could match under either `<ows>`

aengelberg: Oh, I take that back, because `ows` is a regex parser.

daveyarwood: Oh yeah -- I guess the first `ows` would greedily consume all of the whitespace.

daveyarwood:
@aengelberg Do you have any ideas about what might be causing the comment parser to take 300 ms on the Bach cello suite example?

The comment parser is a concatenation of the [comments.bnf](https://github.com/alda-lang/alda/blob/master/server/grammar/comments.bnf) and [clojure.bnf](https://github.com/alda-lang/alda/blob/master/server/grammar/clojure.bnf) grammars.

I ran the score through `(insta/parses comment-parser :partial true)` and there are a ton of steps where it adds `music-data` one character at a time. Could that be part of the issue?

aengelberg:
Out of the 555ms (on my machine) it took to parse the Bach cello suite, about 125ms of that was spent removing comments. I was able to at least shave that off by optimizing the comment parser like so:
```
<non-comment> = #"[^\#\(\)]+"
```
```clojure
;; before change
alda.parser> (time (def parsed (comment-parser (slurp "examples/bach_cello_suite_no_1.alda"))))
"Elapsed time: 126.040464 msecs"
;; after change
alda.parser> (time (def parsed (comment-parser (slurp "examples/bach_cello_suite_no_1.alda"))))
"Elapsed time: 7.186078 msecs"
```

aengelberg:
I was able to take the `score-parser` from 120ms down to 3ms with the following optimization to `score.bnf`:
```
<music-data>      = !calls #"[a-zA-Z]*[^a-zA-Z]*([a-zA-Z][^a-zA-Z]+)*"
```
This takes some thought to fully verify it's doing the right thing. My reasoning is:
- With `!calls`, we guarantee a new `part` is not starting at the current character.
- A `part` always starts with a `name` which would begin with at least two letters.
- Thus, we can safely eat all of the next letters in a row, because if a `part` started after some but not all of the next letters, those letters would have to be `part` of the part name. Hence `[a-zA-Z]*`.
- Now, we can safely eat all of the next non-letters in a row (i.e. numbers, special characters, and whitespace), because a `part` may not start at the current point or in between any of those non-letters. Hence `[^a-zA-Z]*`.
- At this point, we can now eat any letter not followed by another letter, because an `acceptable-name` starts with at least two letters in a row. So we finish the regex off with `([a-zA-Z][^a-zA-Z]+)*`.

Please feel free to correct me if that proof isn't airtight.

A well-known issue with Java regexes would impact this approach: the regex engine will stack-overflow when trying to match thousands of instances of a grouped (`(...)`) expression. Changing to `(?:...)` does not help. Not sure if Alda scores in practice would be affected by this, but that's a definite issue.

The `part-parser` appears to be the primary bottleneck (288ms), so I'll report back if I find any optimizations to that.

daveyarwood:
This helps a lot! These 2 changes alone already makes a huge difference.

It took me a while to wrap my head around your `music-data` regex, but I think I'm getting it now. If I'm understanding correctly, the idea is: instead of reading the part's music-data one character at a time, we read it in chunks, alternating between letters, non-letters, and "one-letter-followed-by-non-letters," and at the end of each pass through these 3 things, we check that the next thing read isn't a `calls`, then repeat. Do I have that right? At any rate, it definitely seems to do the trick! 👍 

It's good to know about the Java regex stack overflow issue... it will be interesting to see if this becomes an issue. I tried to trigger it by making a really long instrument part -- copy-pasting the cello part in the Bach example onto the end of itself -- but ran into "Method code too large!" (see issue #235 ) first. I'll do some more testing once I'm able to resolve that issue, but I think as a general rule, it will be fairly rare for a single instrument part in an Alda score to get big enough to trigger "Method code too large!" on its own, let alone to trigger the regex group stack overflow. I think in the near future, the recommended score-writing method is going to be to define a bunch of sequences of notes (outside of parts) and then compose the instrument parts out of the defined sequences by referencing variables ( #131 ), so the use case for having a super long instrument part will become less and less.

I really appreciate all the help! I'm excited to see if you find anything else we can optimize in `part-parser`.

PR #244, opened 2016-07-07
Clojure plugins

rwatts07:
This PR adds some basic functionality for user-defined clojure plugins to manipulate code before it is passed to the parser.

`/resources/plugins.clj` should define a function `apply-clojure-plugin` that takes alda code as input.  The result is then passed through the rest of the parsing process.  For a simple test case, I've created a plugin that replaces the shorthand `Em` with an E minor chord.  A more useful case might be defining a guitar tab syntax that converts guitar tabs into alda code.  In general, any text manipulation that returns valid alda code should be fine.

In the future it would be nice to add a bit more flexibility -- different languages, grabbing plugin code from github repos, etc.  Since this is my first time contributing to alda, I thought I'd keep it short for now, but hopefully I can work on adding some of these features in the future.


daveyarwood:
This looks great, thanks for the PR!

I'm intrigued by this -- it's definitely a nice first step toward a robust plugin system. But I'm not sure if this would actually work in the way that we would want it to work. I think we would want Clojure programmers to be able to write a Clojure program and include Alda as a dependency, and also include a `plugins.clj` on their resource path that defines an `apply-clojure-plugin` function, and our intent is for Alda to be able to read and evaluate that file, and apply the function to the parser's input before parsing.

I briefly tried to play around with this in a sample Clojure project, and I wasn't able to get it to recognize the plugin (I copied `resources/plugins.clj` from this branch into the `resources/` folder of my project, and made sure that that folder was on the resource path). It makes me think that maybe Alda, when used as a dependency, doesn't have dynamic access to the resource path of the project that is using it as a dependency.

It's entirely possible that I'm overlooking something crucial... have you been able to get it to work in a sample project?

If the resource-path route doesn't work for this, I have a feeling it may be easier to take a stab at reading files from a GitHub repo.

rwatts07:
> It's entirely possible that I'm overlooking something crucial... have you been able to get it to work in a sample project?

I haven't tried including alda as a dependency in a clojure project, no.  As written, `plugins.clj` is assumed to be on the classpath of the alda project, although I would imagine it could be modified in that respect. 

Could you describe the dependency use case a bit more?  It sounds a bit different than what I usually picture as "plugins" but maybe I'm misunderstanding. For example, if I want to be able to parse guitar tabs in alda, would I create a separate clojure project with alda as a dependency?

daveyarwood:
Oh, I see -- so is your use case that to apply a plugin, you would clone the Alda repo, and add a `resources/plugin.clj` file, then build an alternate version of the `alda` executable where it applies the plugin?

I was assuming this was for people who want to use Alda as a dependency in a Clojure project -- i.e. you want to write a Clojure program that parses Alda code, and you want to modify the parsing behavior by including a plugin on the classpath. That strikes me as kind of a neat use case, but I'm not sure if it works because as far as I can tell, you can't modify the classpath of a dependency jar.

rwatts07:
> Oh, I see -- so is your use case that to apply a plugin, you would clone the Alda repo, and add a resources/plugin.clj file, then build an alternate version of the alda executable where it applies the plugin?

That was originally my use case -- I wanted a way to write a clojure program that took guitar tabs and translated them into alda before passing them along to the parser.  Now that I'm thinking about it though, it would certainly be best to not have to clone/build the project for every plugin.  

Maybe having a path (configured via an environment variable) that alda checks for `plugins.clj`?  Since `load` happens at runtime that should work as long as you can get the file, right?

daveyarwood:
Yeah, I like that approach. We could default to `~/.alda` if the environment variable isn't set.

I think what I'd ultimately like to do is set up the Alda parser to read lines at the beginning of a file that specify which plugins to use -- for example, including the line `plugin rwatts07/guitar-tab "0.1.0"` would clone that GitHub repo, checkout the "0.1.0" tag, and copy it to `$ALDA_PATH/plugins/rwatts07/guitar-tab/0.1.0`, and then load `plugins.clj` in that directory apply the plugin. On subsequent parses, the parser would see that the directory already exists and skip straight to loading and applying the plugin.

daveyarwood:
@rwatts07 Thanks again for the PR -- it's exciting to think through this stuff.

Hope you don't mind if I close it though, as we will be taking a different approach to implementing plugins.

PR #250, opened 2016-07-22
Update CHANGELOG.md

elyisgreat:


daveyarwood:
Whoops... so I was on a plane when I wrote the code for the 1.0.0-rc24 release, and it turns out that I forgot to push my changes to master. 😓   This included updating the changelog, so I at least have the main "changes" part in the changelog now. Your PR also includes a note about undefined variables throwing an error now, as a breaking change -- that's a good idea, I'll add it on master and ghost-commit it as you :)

Thanks for the PR!

elyisgreat:
also needs to be updated on the [releases](https://github.com/alda-lang/alda/releases) page

daveyarwood:
Good call. Done!

Issue #259, opened 2016-08-29
midi output to non-java device


0atman:
While there are some excellent soundfonts out there, and thanks to this project I am now aware of them, nothing can beat using your own DAW, VSTs or physical devices for output.

This feature would also give us MIDI file saving, in a round about way ("record" alda into any DAW, then save the midi to file from there)

daveyarwood:
Great idea! I think this could potentially be a very easy feature to implement.

We are already firing MIDI note on/off events (as well as volume and panning messages and assigning instrument patches to channels) in real time, using our own scheduler. We're just sending these MIDI messages to a JVM MIDI synthesizer instance, instead of emitting them in a way that they could be used by another program.

We could add an option that would tell Alda to just emit the messages instead of sending them to the JVM MIDI synth.

0atman:
I'm none too familiar with the JVM or clojure, but where could I get started on this?

daveyarwood:
That would be great! I'll be happy to help with any questions/problems you run into, including help with Clojure. I realize not a lot of people speak Clojure, but I want to do what I can to help change that :)

So here are my thoughts so far, and hopefully some of this is helpful:

I think the easiest way to start on this would be to just modify the code we have so that it's emitting the MIDI messages instead of playing them in the JVM, and get that setup working locally where Alda is emitting MIDI messages and your DAW is receiving them. (Then we can tweak it so that we still get the current behavior by default, but we can enable this new behavior via a CLI option -- we can worry about that part later.)

I don't know a whole lot about transmitting MIDI messages with Java yet, but from what I can tell, the first thing we need to figure out is how to choose the MIDI output device. I think the most flexible way would be to let the user tell us which device to use. We could make it so that the user runs `alda list-devices` or something and we show the available devices, then the user can run something like `alda --device 5 play -f my-score.alda`, and Alda would know to use MIDI device "5" (assuming they're numbered consistently?) to transmit MIDI messages instead of using its internal MIDI synth. (We can also worry about this part later -- to start, maybe you could just find the device you need and hard-code it in.)

---

Some resources about transmitting MIDI messages in Java:

* https://docs.oracle.com/javase/tutorial/sound/MIDI-messages.html
* https://docs.oracle.com/javase/7/docs/api/javax/sound/midi/MidiDevice.html
* https://docs.oracle.com/javase/7/docs/api/javax/sound/midi/Receiver.html
[...]

PR #263, opened 2016-09-06
[WIP] Adding direct MIDI output

0atman:
From https://github.com/alda-lang/alda/issues/259.

> While there are some excellent soundfonts out there [...] nothing can beat using your own DAW, VSTs or physical devices for output.
> 
> This feature would also give us MIDI file saving, in a round about way ("record" alda into any DAW, then save the midi to file from there)

Currently a work in progress, most likely only of interest to @0atman, @daveyarwood and, potentially, nobody! Watch this space...

### DIFF
--- a/server/src/alda/now.clj
+++ b/server/src/alda/now.clj
@@ -93,8 +93,9 @@
    evaluating alda.lisp code, e.g. (score (part 'piano' (note (pitch :c)))),
    or an atom referencing such a map."
   [score & [play-opts]]
-  (sound/with-play-opts (merge {:async? true} play-opts)
-    (sound/play! (if (instance? clojure.lang.Atom score)
-                   @score
-                   score))))
+  (println "testing alda dev setup"))
+  ;(sound/with-play-opts (merge {:async? true} play-opts)
+  ;  (sound/play! (if (instance? clojure.lang.Atom score)
+  ;                 @score
+  ;                 score))))
 


server/src/alda/now.clj
```diff
@@ -93,8 +93,9 @@
    evaluating alda.lisp code, e.g. (score (part 'piano' (note (pitch :c)))),
    or an atom referencing such a map."
   [score & [play-opts]]
-  (sound/with-play-opts (merge {:async? true} play-opts)
-    (sound/play! (if (instance? clojure.lang.Atom score)
-                   @score
-                   score))))
+  (println "testing alda dev setup"))
```
0atman: First problem, this little test, suggested by Dave, doesn't seem to do anything.

0atman: Wait, maybe it's my machine, let me test on my work machine tomorrow!

daveyarwood: I can explain what's going on here -- since 1.0.0-rc35, the server doesn't execute that part of the code (the "playing audio") part anymore. Now we have separate worker processes that handle parsing, evaluating and playing Alda scores.

Now that there are more moving parts, the setup is slightly more complicated, but hopefully not too much so... I added [some info](https://github.com/alda-lang/alda/blob/master/doc/development-guide.md#alda-worker) to the development guide about starting server and worker processes for development.

There are step by step instructions in the link above, but basically, the new way to do it is to:

- Start a development server with the `ALDA_DISABLE_SUPERVISOR` environment variable set (this keeps the server from inadvertently killing your debug worker process) and the `--workers 0` option so it doesn't start any worker processes of its own.
- Start a development worker process with `ALDA_DEBUG` environment variable set and the `--port 12345` option, where `12345` is the _backend port_ where the server communicates with its workers. (The server logs that port number out when it starts.)
- If you have this working correctly, you'll start to see the worker logging "Got HEARTBEAT from server" about once per second. Then when you send a request to the server by running e.g. `alda play -c 'piano: c'`, the server will forward the request to your worker (because it's the only one available) and you should see `testing alda dev setup` printed in the output of the worker.

0atman: Ah I see! Makes sense, testing now, thank you.

0atman: It works! 3 moving parts ain't so bad! Thank you, I'll get deving.

2016-09-06-233701_1334x765_scrot

0atman:
Been too busy to work on this recently, will come back to it asap

daveyarwood:
@0atman I'm going to close this PR for now, in the interest of keeping things tidy. But please feel free to re-open it if you find the time to pick this back up!

PR #264, opened 2016-09-14
Detect suspension, cycle workers

### LINKED ISSUE #160: MIDI audio is delayed after suspending the server
I get pretty much immediate responses from the server\* after first running it, but after leaving the server running for a while (e.g. if I leave my computer and come back hours later), it may take several seconds or more before I hear anything. Restarting the server fixes it.

\*meaning, if I run `alda play -r -c 'piano: c'`, I hear the note immediately

No idea what the issue is at this point, but figured I would report it in case anyone has any ideas.

daveyarwood:
Closes #160.

### DIFF
--- a/server/src/alda/server.clj
+++ b/server/src/alda/server.clj
@@ -38,12 +38,25 @@
                             (remove #(contains? @worker-blacklist (:address %)))
                             doall))
 
+(defn- friendly-id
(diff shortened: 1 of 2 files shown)


daveyarwood:
:shipit:

PR #272, opened 2016-09-22
Update attributes.md

wmwragg:
I believe that the (G natural) was missing the `_` accidental. I'm new to Alda, and just reading the docs to get familiar with it, so I may be incorrect here.


daveyarwood:
You're right. Thanks for the PR!

PR #273, opened 2016-09-22
Update voices.md

wmwragg:
I've just started learning Alda, and was a little confused reading this, about how you actually used the voices with an instrument. Once I checked some of the examples it became clearer, but I think adding an instrument to the example makes it much clearer.


daveyarwood:
You're right, I could see how that would be confusing. Thanks!

PR #274, opened 2016-09-22
Update attributes.md

elyisgreat:
Added in the bit about `set-duration` instead of `duration`


daveyarwood:
Thanks for the PR! I like the idea of explaining this. Could we also include an example? It may not be intuitive that `set-duration` accepts a number of beats, and not a note length.

elyisgreat:
Sure! Feel free to add that in.

Issue #269, opened 2016-09-19
Optimize JVM options to minimize memory use, etc.


daveyarwood:
Because the server and workers are all separate processes with their own JVMs, it would be good if we took steps to make sure that the processes are as lightweight as possible while still having the resources they need.

Currently we are not using any JVM options, except for Clojure direct linking. We could add in some options to set a cap on memory use, heap space, etc.

Alda's JVM options are set [here](https://github.com/alda-lang/alda/blob/master/build.boot#L57) in build.boot. These options are used in the executable when it is built. They are shared between the client, worker and server processes, since they share the same jar file / executable.

I could use some help with this -- I don't have a great deal of experience with tuning JVM applications.

feldoh:
I'm up for having a go. Given that this is a music app I'm assuming you want to prioritise minimal pause times as GC in the middle of a melody isn't ideal. Also in terms of heap size I'm assuming it's going to be relatively small but I can't choose a collector appropriately or indeed heap limits without a fairly worst case sample. Can you give an example of a particularly heavy workload. I'm guessing either CMS or G1 will be best (although if you really want low resource usage the throughput collector would actually be best, but I'd never advise that for an interactive system) I'd prefer to benchmark a worst case rather than just guess though ;) One thing to consider is to reduce pauses it is generally advised to assign minimum to your estimated maximum to avoid heap resizing which is an expensive operation but that is in direct opposition to keeping the resource usage down. Especially if you consider something like using the cli options to play a single note ending up taking 4G or something silly. Given that I'd be tempted to start low and take the resizing hits. So given the wide load variety I'm not sure which situation you want to optimise for. Give me some pointers as to your ideal trade-offs and I'll have a crack.

daveyarwood:
Awesome, thanks for offering to help! I'll be curious to see what you find is most resource-efficient.

I also feel like starting low might yield the best results. Basically, I think right now, each worker process is probably given more resources than it needs, so it would be interesting to see what are the lowest caps we can set without performance being negatively impacted. It seems like an Alda worker process should be able to be pretty lightweight, so we could allow the user to scale up "horizontally" by adding more workers.

Signs of performance impact to watch for would be things like:

- audio stuttering or cutting out
- notes delayed / rhythms getting skewed
- a score otherwise "not sounding like it should"
- lack of immediate response from server/workers

[...]

PR #276, opened 2016-10-05
#269 Use G1 and add basic memory limits

feldoh:
Empirically G1 was found to put the least load on the system on the three boxes tested.
Memory usage between 250m and 800m.
CPU usage hovers around 9% of one core on my test machines VS 30 with previous values.

### DIFF
--- a/build.boot
+++ b/build.boot
@@ -54,7 +54,7 @@
       (format "%s.999" n)
       version)))
 
-(def jvm-opts #{"-XX:+UseG1GC -XX:MaxGCPauseMillis=100 -Xms256m -Xmx1024m -Dclojure.compiler.direct-linking=true"})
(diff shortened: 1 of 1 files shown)


feldoh:
You may also like to try out -XX:InitiatingHeapOccupancyPercent=0 which turns on continuous collection. Oddly I found this lowered overall load, but I'm not sure I believe that figure. Without lots of garbage that doesn't make much sense.

daveyarwood:
Thanks for the PR!

It should work the way you have it (coincidentally), but would you mind changing it so that each JVM option is a separate string? i.e. `#{"-XX:+UseG1GC" "-XX:MaxGCPauseMillis=100" "-Xms256m" "-Xmx1024m" "-Dclojure.compiler.direct-linking=true"}`

feldoh:
I'd actually managed to not notice that was a set :p

daveyarwood:
Trying this out locally, it seems to work great! CPU usage is definitely improved from what's on master at the moment, even with 4 workers. Excellent work!

PR #278, opened 2016-10-11
Separate jeromq-sanity-check zcontext tests from jeromq-sanity-check ctx tests

benoid:
If the zcontext tests failed, the scripts would never get to the ctx tests.  I put the zcontext and ctx tests in separate files.

### DIFF
--- a/scripts/jeromq-sanity-check-client-ctx.boot
+++ b/scripts/jeromq-sanity-check-client-ctx.boot
@@ -16,12 +16,11 @@
   [socket]
   (println "Hello world client/server example:")
   (dotimes [n 5]
-    (let [req (.recv socket)
-          res (format "Hello #%s from server" (inc n))]
-      (println "Received request:" (String. req))
-      (Thread/sleep 1000) ; simulate doing work
-      (println "Sending response:" res)
-      (.send socket res))))
+    (let [req (format "Hello #%s from client" (inc n))]
+      (println "Sending request:" req)
+      (.send socket req))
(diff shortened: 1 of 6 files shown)


daveyarwood:
Nice, this is a good idea.

We could also simplify the scripts if each script is only dealing with one type of context. We can get rid of the SocketMaker and Destroyable protocols that I made just to differentiate between the two types of contexts, and just use:

- `(.createSocket ctx socket-type)` and `(.destroy ctx)` in the ZContext script
- `(.socket ctx socket-type)` and `(.term ctx)` in the Ctx script

benoid:
I am not sure I will be able to do the simplifications until we diagnose the jeromq issue affecting my machine, because I'll have no way of knowing whether the script works or not.

daveyarwood:
I can do the script simplifications real quick after merging to master -- I think it would be good for these scripts to be as close as possible to the Java examples in the zguide repo.

Thanks again for the PR!

benoid:
Before you do that, I just made those changes. I tested them, and they work.  Do you want me to push them before you merge?

benoid:
And by they work, I mean they are actually communicating with each other!

daveyarwood:
Oh, whoops -- I just made the changes myself on master :)

That's awesome that the scripts are working! Onward!

benoid:
I can confirm that the changes you made also work on my system.

daveyarwood:
OK, excellent 👍

PR #279, opened 2016-10-11
update sanity-check example output

benoid:
Update the example output to match the new file names.


daveyarwood:
Ah, good catch. Thanks!

Issue #242, opened 2016-06-20
Java client code for printing messages/errors is buggy


daveyarwood:
Full disclosure: I'm not the best Java programmer. When I wrote the Java code for the Alda client, I included a couple of convenience functions for printing messages and errors that bubble up from the Alda servers. These functions are basically wrappers around `printf`, with some program-specific context like the port the server is running on.

Occasionally, things can get weird if the message or error strings happens to contain some unexpected formatting characters like `%s`. This may be what was happening [here](https://github.com/alda-lang/alda/issues/241#issuecomment-227047730) in the stacktrace @elyisgreat was getting.

I'm opening this up to all the Java programmers out there:
1. Is there a way [these convenience functions](https://github.com/alda-lang/alda/blob/master/client/src/alda/AldaServer.java#L97-L119) can be made more robust so that they won't fail if the message string contains unexpected format characters?
2. Is the `java.util.UnknownFormatConversionException: Conversion = '1'` error indicative of some other problem?

jgkamat:
Sorry for not contributing in a long time, I've been kind of backed up on other projects, but I'll probably work  a bit more on alda in about a month!

That error message is likely something wrong with the inputs into printf, either `%` directives that don't exist (like `%q` I think?), or perhaps there are an incorrect number of `%` directives. From the error, I'm guessing that there is a `%1` in your input to printf.

In general, java programmers don't use printf, as the built in string addition is more robust and easier to use.

The possible solutions we have are:
1. (the easiest) Make sure inputs into the `msg` functions are valid printf statements, and make sure you have the correct number of arguments. Effectively, you would be treating `msg` as a printf function, as you do now. If you do this, it's up to callers of this function to make sure the input is valid. In this case, you would have to make sure the `prefix` variable does not contain any printf special characters.
2. Switch over to use string addition entirely, by passing in a list of args, and your function will string concatenate them and print them out.
3. (probably the cleanest) pass a pure string into `msg` and let the caller deal with concatenation and message formatting. The caller could make a sprintf (equivalent) call themselves.

daveyarwood:
Fixed in [1.0.0-rc48](https://github.com/alda-lang/alda/releases/tag/1.0.0-rc48).

PR #285, opened 2016-11-12
Issue 242: Remove variadic args to .msg method and replace printf with println

bbqbaron:
This is for [Issue 242](https://github.com/alda-lang/alda/issues/242). It looks like there were only two actual uses of the `...args` for message formatting, so I just called `String.format` outside `msg` and then removed the extra arguments entirely. It seemed pretty simple, so I may be missing some context here!

### DIFF
--- a/client/src/alda/AldaServer.java
+++ b/client/src/alda/AldaServer.java
@@ -44,7 +44,7 @@ public class AldaServer extends AldaProcess {
     }
   }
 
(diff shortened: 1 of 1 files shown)


daveyarwood:
This looks good - I think it will fix #242. The real problem there was using`printf` inside of the `msg` function, where we don't have full control over the string that's being formatted. This approach is more sensible.

Thanks a lot!

PR #292, opened 2016-11-19
Remove explicit voice group event from parse tree

bbqbaron:
From [286](https://github.com/alda-lang/alda/pull/286), cleaning up after myself by removing the `voices` element from the parser tree entirely. It turned out to be as simple as just swallowing the token with `<>`, deleting the handling code for it, and removing it from assertions. The one thing that `:voice-group` does, `assoc :voice-instruments {}`, already seems to be handled by the events for individual voices.

This is intended to cause no behavioral changes in Alda; pure code cleanup.


bbqbaron:
Whoops! I seem to have missed the major reorg on master. I'll update and resubmit.

daveyarwood:
@bbqbaron Sorry for the switcharoo! :)

PR #302, opened 2018-02-02
Update README.md - Added server start step to Demo section

TBuc:
Added server start step to Demo section


daveyarwood:
Great idea! I wrote that part of the README back in the day back before we had the client/server architecture.

Thanks for the PR! :+1:

PR #303, opened 2018-02-04
Update notes.md

TBuc:
Replace broken link

### DIFF
--- a/doc/notes.md
+++ b/doc/notes.md
@@ -1,6 +1,6 @@
 # Notes
 
-Alda's syntax for **notes** is heavily inspired by [MML](http://www.nullsleep.com/treasure/mck_guide).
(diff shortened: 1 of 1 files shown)


daveyarwood:
Ah, it's a shame that that link went down! That must have happened recently.

I think it's a good idea to link to the Wikipedia article, though, at any rate.

Specifically, we should link to the "Modern MML" section here, as that is the syntax that inspired Alda's syntax: https://en.wikipedia.org/wiki/Music_Macro_Language#Modern_MML

TBuc:
Yep! Hopefully Wikipedia's entry will stay up a little bit longer than other websites :-)
Modified as requested.

daveyarwood:
Awesome, thanks for the PR!

PR #353, opened 2021-05-01
Set up infrastructure for MusicXML importer and handling of basic tags

Scowluga:
This PR sets up the MusicXML importer in an experimental state
- There is no code generation, so the actual output is just a bootstrapped copy of the `play` command
- Only basic versions of notes, rests, and key signatures are currently supported

The `import` command is marked as hidden while experimental. Subsequent PRs will add additional handling for tags


client/cmd/import.go
```diff
@@ -0,0 +1,255 @@
...
+				aurora.BrightYellow(importFormat),
+				aurora.BrightYellow("musicxml"),
+			)
+		}
+
+		// TODO (experimental): remove warning log
+		log.Warn().Err(help.UserFacingErrorf(
+			`The %s command is currently experimental. Imported scores may be incorrect and lack information.`,
+			aurora.BrightYellow("import"),
+		)).Msg("Experimental command")
```
daveyarwood, line 91: It would be simpler to log a string here instead of an error, something like:

```go
log.Warn().
  Str(
   "detail",
   fmt.Sprintf(
     `The %s command is currently experimental. Imported scores may be incorrect and lack information.`,
     aurora.BrightYellow("import"))).
  Msg("Experimental command")
```

Or maybe even:

```go
log.Warn().
  Msg(
   fmt.Sprintf(
     `The %s command is currently experimental. Imported scores may be incorrect and lack information.`,
     aurora.BrightYellow("import")))
```

client/cmd/import.go
```diff
@@ -0,0 +1,255 @@
...
+				return err
+			}
+
+			code_generator.Generate(scoreUpdates, file)
+
+			fmt.Fprintf(os.Stderr, "Imported score to %s\n", outputAldaFilename)
+			if err := file.Close(); err != nil {
+				return err
+			}
+			//return nil
```
daveyarwood, line 140:
```suggestion
```

daveyarwood (approved):
Looks good overall! 

@Scowluga I've left a bit of feedback that you can take or leave. Feel free to merge whenever you're ready!

client/cmd/import.go
Scowluga: Resolved.

client/cmd/import.go
Scowluga: Great this is a lot better. Just defaulted to creating an error. Resolved.

Scowluga:
@daveyarwood Great thanks for the feedback. I've resolved both of them and rebased. Merging now.

PR #356, opened 2021-05-31
MusicXML Importer - Refactor and Handling of Repeats

Scowluga:
## Refactor
* Dealing with repeats, I now need to account for 3 new `model.ScoreUpdate` elements that have nested slices: `Repeat, OnRepetitions, EventSequence`. 
* This means I can no longer naively brute force the recursive aspect of modifying `[]model.ScoreUpdate` as I did when there were only chords. 
* This PR completes a significant refactor of the importer to encapsulate `[]model.ScoreUpdate` modification. Now all `elementHandler` functions can interface with the `musicXMLImporter` in a standard way. All recursion is managed in `importer.go` (see `modifyAt` and `findLast`). All recursive properties of Alda IR are encapsulated in `util.go` under `getNestedUpdates` and `modifyNestedUpdates`. 
* This refactor was quite tedious, but it sets a solid flexible and extendible foundation to build upon. Future handlers will be simpler to write and can rely on this interface. 

## Repeats 
* This PR adds handling for repeats. 
* Repeats are quite complicated, as we have to consider multiple endings. 
* Furthermore, since Alda `model.Note` elements do not track octaves, we must maintain the integrity of the importer's current octave through repetitions and endings. 
	* To do this I've added the `sectionStartOctave` and `endingStartOctave` parameters to the `musicXMLVoice`

## Documentation 
* Created an `interop/musicxml/doc` folder and moved in `interfacing-with-musicxml.adoc`. 
* Created a doc explaining how to add MusicXML test cases.


daveyarwood (approved):
Looks good! Feel free to merge whenever you're ready.

PR #358, opened 2021-06-05
Add support for dynamic markings 

Scowluga:
Alda supports volume setting via `(vol __)` and `(vol! __)` attribute updates. Many scores do not necessarily have numeric volumes, but instead have dynamic markings. This PR introduces basic dynamic markings into Alda via Lisp attribute updates. 

## Dynamic Markings
This PR will introduce the `(mf)` and `(mp)` dynamic markings, as well as the `(pppppp)`, ..., `(p)` and `(ffffff)`, ..., `(f)` markings. We choose to support up to 6 pianos or fortes following the MusicXML standard [here](https://www.w3.org/2021/06/musicxml40/musicxml-reference/elements/dynamics/). 

This PR does not add support for more complex dynamics such as fortepiano or sforzando. We believe this would add unnecessary complication into the Alda language. 

## Dynamic Volumes
A MIDI note's volume is represented by velocity which is an integer in the range `[0, 127]`. The majority of existing software maps dynamics to velocities following a linear scale. See the Wiki page [here](https://en.wikipedia.org/wiki/Dynamics_(music)#Interpretation_by_notation_programs). Alda will follow a similar uniform distribution, but supporting up to 6 pianos and fortes instead of the usual 3. 

Alda internal representation uses a float in the range `[0, 1]` instead of velocity. I have added a mapping with velocities commented. 

## Default Dynamic
With the introduction of dynamics, we will also change every Alda score to start with a default dynamic. We select `mf` as the default dynamic following MuseScore as explained [here](https://musescore.org/en/handbook/3/dynamics#:~:text=The%20default%20dynamic%20of%20the,than%20or%20greater%20than%20this.).

### DIFF
--- a/client/model/attributes_test.go
+++ b/client/model/attributes_test.go
@@ -331,7 +331,7 @@ func TestAttributes(t *testing.T) {
 				PartDeclaration{Names: []string{"piano"}},
 			},
 			expectations: []scoreUpdateExpectation{
-				expectPartVolume("piano", 1.0),
+				expectPartVolume("piano", DynamicVolumes["mf"]),
 			},
 		},
 		scoreUpdateTestCase{
@@ -364,7 +364,7 @@ func TestAttributes(t *testing.T) {
 				AttributeUpdate{PartUpdate: DynamicMarking{Marking: "ppp"}},
 			},
 			expectations: []scoreUpdateExpectation{
-				expectPartVolume("piano", 0.23683),
+				expectPartVolume("piano", DynamicVolumes["ppp"]),
 			},
 		},
 		scoreUpdateTestCase{
--- a/client/model/attributes_test.go
+++ b/client/model/attributes_test.go
@@ -358,6 +358,16 @@ func TestAttributes(t *testing.T) {
 			},
 		},
 		scoreUpdateTestCase{
+			label: "set volume using dynamic marking",
+			updates: []ScoreUpdate{
+				PartDeclaration{Names: []string{"piano"}},
+				AttributeUpdate{PartUpdate: DynamicMarking{Marking: "ppp"}},
+			},
+			expectations: []scoreUpdateExpectation{
+				expectPartVolume("piano", 0.23683),
+			},
+		},
+		scoreUpdateTestCase{
 			label: "initial track volume",
 			updates: []ScoreUpdate{
 				PartDeclaration{Names: []string{"piano"}},
--- a/client/model/attributes.go
+++ b/client/model/attributes.go
@@ -334,6 +334,35 @@ func (tvs TrackVolumeSet) updatePart(part *Part, globalUpdate bool) {
 	part.TrackVolume = tvs.TrackVolume
 }
 
+var DynamicVolumes map[string]float64
+
+func init() {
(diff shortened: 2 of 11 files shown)


client/model/attributes_test.go
```diff
@@ -357,6 +357,16 @@ func TestAttributes(t *testing.T) {
...
 			},
 		},
+		scoreUpdateTestCase{
+			label: "set volume using dynamic marking",
+			updates: []ScoreUpdate{
+				PartDeclaration{Names: []string{"piano"}},
+				AttributeUpdate{PartUpdate: DynamicMarking{Marking: "ppp"}},
+			},
+			expectations: []scoreUpdateExpectation{
+				expectPartVolume("piano", 0.23683),
```
daveyarwood, line 367: I think it might be good to define this in terms of the DynamicVolumes map, just so that if we ever update it, we don't have to do it in two places:

```suggestion
				expectPartVolume("piano", DynamicVolumes["ppp"]),
```

daveyarwood (approved):
Looks good!

I think it would also be a good idea to have a look at the scores in the `examples` directory, make sure that they sound OK, and perhaps change some of the `vol` attributes to dynamic markings, so that we have more test coverage. (If you didn't notice this already, `client/parser/examples_test.go` does some smoke tests where it parses and evaluates every score in the `examples` directory. I'm also in the habit of listening through every score before I cut a release, so that I can see if my ears pick up on anything that might be a regression.)

It might also be a good idea to add a new example score that makes use of dynamics. It could just be something really simple, like:

```alda
electric-guitar-distorted:
  o2
  (pppppp) e
  (ppppp) e
  # etc.
  (fffff) e
  (ffffff) e
```

client/model/attributes_test.go
Scowluga: Changed this. Also ran the tests and realized there's a test for "initial volume" equal to 1.0 that fails. Not sure how I missed that. I updated that to use `DynamicVolumes["mf"]`.

Scowluga:
Good point, I hadn't thought much into the examples. I've added `dynamics.alda` to showcase all the existing dynamics to make sure we can parse without issue. Note that `(pppppp)` corresponds to a velocity of 1, and I can't even hear it at max volume. But velocity of 1 theoretically makes a sound. 

I've also tried to update the existing examples the best I can to account for dynamics. Most do not set volumes, in which they are now just overall more quiet. Some set volumes, but also set volumes at the start. For `debussy_quartet.alda` I switched the volume updates to use dynamics, but I also started each part at `(ff)` as otherwise they would start at `(mf)` which is incorrect.

Scowluga:
I've updated documentation. See: 
* https://github.com/alda-lang/alda/blob/feature/dynamics/doc/alda-2-migration-guide.md#score-starting-volumes
* https://github.com/alda-lang/alda/blob/feature/dynamics/doc/attributes.md#dynamic-markings

This is just what I imagine we might want to do. Let me know how I should change it or feel free to just update it yourself. Let me know.

daveyarwood:
This is fantastic. Thank you! :100:

PR #360, opened 2021-06-16
MusicXML Importer - Dynamics

Scowluga:
With the merging of https://github.com/alda-lang/alda/pull/358 (Dynamic markings in Alda), we can now approach dynamics in MusicXML import. 

In particular, we can import the "easy" dynamics directly as implemented by the Alda language. For more difficult dynamics like sforzando or fortepiano, we do not plan to support these due to complexity. We'll just log a warning to the user. 

## Using Recurse Handler by Default
This PR updates the importer to recurse on elements by default instead of skipping them. This makes more sense as we'll just recurse, ignoring the elements we don't support, and importing the ones we do. 

I originally intended to skip by default in order to force developers to specify usage of recurse handler for interesting tags. But this is not scalable given the scope of MusicXML. Recursing by default is more flexible and sustainable.


daveyarwood (approved):
Makes sense :+1:

PR #361, opened 2021-06-21
Start of Alda Decompiler (Generator & Formatter)

Scowluga:
The need for an Alda decompiler is twofold: 
1. With the start of interoperability, we are now able to import MusicXML into Alda IR (`ScoreUpdate`). Although we can do lots of stuff from this representation like output to JSON or play, we can't actually obtain Alda code itself from it. The decompiler will serve to decompile `ScoreUpdate` structs to Alda code so users can start writing Alda from imported music. 
2. Many great languages come with great formatters (ex `gofmt`). Alda may eventually aim to support built-in formatting. This is most easily done by scanning to `parser.Token` then formatting back to code. This PR will set up this formatter in the most basic form. 

This PR sets up both the generator and formatter with simple / partial implementations.


client/cmd/import.go
```diff
@@ -1,7 +1,8 @@
 package cmd
 
 import (
-	"alda.io/client/code-generator"
+	"alda.io/client/formatter"
+	"alda.io/client/generator"
```
daveyarwood, line 5: I think I'd prefer that we call these `code_formatter` and `code_generator`, so that the package names make it very obvious what they do.

client/cmd/import.go
```diff
@@ -121,16 +122,18 @@ Currently, the only supported output format is %s.`,
 			return err
 		}
 
+		tokens := generator.Generate(scoreUpdates)
```
daveyarwood, line 125: Capturing our conversation on Slack:

We would like to break this apart a bit, so that the code generator produces an AST instead of a flat list of tokens. Along with that, we'd also like to make it so that the _parser_ produces an actual AST instead of a flat list of tokens. We are sort of already doing this, but it's entangled with the part where we turn the list of tokens into a list of ScoreUpdates. So one thing we'd like to do is refactor it so that there is an explicit step where we turn the list of tokens into an AST, and then a subsequent step where we turn the AST into a list of ScoreUpdates.

We would also implement the reverse, for the code generator: "unparsing" a list of ScoreUpdates into an AST.

Then the formatter would operate on the AST, turning it into a nicely formatted string of Alda code.

daveyarwood (commented):
Leaving a couple of comments here while our conversation on Slack is still fresh on my mind.

This is as much for my own reference as it is for yours :)

I'll take a look at doing the refactor to add the "tokens => AST" step in the parser soon, after I finish releasing Alda 2!

client/cmd/import.go
Scowluga: Yeah `generator` especially is a little ambiguous, I agree.

Scowluga:
Closing this pull request to split it into 3 PRs: 
1. Code generator
2. Code formatter
3. Usage in MusicXML importer

The discussion here will remain if we want to reference it.

PR #364, opened 2021-06-27
MusicXML Importer - Optimization

Scowluga:
Due to the differences between MusicXML and Alda representation, while importing we must maintain some extra information that is not present in idiomatic Alda. We want to apply post processing on the importer state to remove redundant information and generate more idiomatic Alda. 

This PR adds a post processing step in the MusicXML importer with the following key method: 
```
func (processor *postProcessor) processAll(
	updates []model.ScoreUpdate,
) []model.ScoreUpdate {
	// We standardize barlines in post processing to improve duration removal
	updates = standardizeBarlines(updates)
	updates = processor.removeRedundantAccidentals(updates)
	updates = processor.removeRedundantDurations(updates)
	return updates
}
```

## Standardize Barlines 
Barlines are introduced into score updates with duration components (rest, note). This is because we might have to deal with ties, and follows the Alda convention. However after all is said and done, we will have many updates that have barlines as the last element in their duration components. Then the barline can just as easily be extracted. This extraction process is not necessary, but it creates equivalent Alda and serves to make testing and removing durations easier. So we do it. 

## Remove Redundant Accidentals 
In Alda, a key signature automatically applies accidentals to each note. In MusicXML, the key signature is just visual, and each note must have the correct pitch. So in the importer Alda there are many redundant accidentals that should be removed. 

However we shouldn't just naively remove all accidentals. In normal sheet music, after a pitch has diverged from it's normal state as defined by a key signature, when it returns back to the key signature it should have accidentals to re-iterate this return. For example, in G major if we have an F natural followed by an F sharp, that second F should have the sharp, even if in Alda it is not necessary since it would be automatically applied by the key signature. 

So in this step we do some extra work by maintaining `currentNoteState` which represents if a note has diverged from its key signature. 

## Remove Redundant Durations 
The most difficult of all is removing redundant durations. In Alda a note/rest can have `nil` durations, and this will simply copy the last encountered duration. 

Doing this in post processing is not as easy as it seems. Originally I just processed the `[]model.ScoreUpdate`s linearly and maintained a duration. But it turns out this does not work since we have repeats. To properly remove durations we would have to follow the execution of the score itself, which is non-trivial. 

Instead I have applied a heuristic in removing durations. I only remove durations that are super unnecessary. These are durations that happen within the same measure, and are not at different nesting levels (chords, repeats). This leads to Alda code that does not have every possible duration removed, but has most unnecessary ones removed. I think this is quite nice, and the resulting test Alda code looks very readable.  

Essentially, to remove all possible durations is very difficult. But removing this subset is very doable, and leads to Alda code that I'd argue is more readable. So I've gone ahead and done that.


Scowluga:
Code for `removeRedundantDurations` for archival purposes. This works and does what it is written to do, but it's not correct. When faced with repeats, it breaks down completely. 
```
func removeRepeatedDurations(
	update model.ScoreUpdate, processor *postProcessor,
) model.ScoreUpdate {
	getDurationComponents := func(
		update model.ScoreUpdate,
	) []model.DurationComponent {
		if reflect.TypeOf(update) == noteType {
			return update.(model.Note).Duration.Components
		} else if reflect.TypeOf(update) == restType {
			return update.(model.Rest).Duration.Components
		} else {
			return nil
		}
	}

	setDurationComponents := func(
		update model.ScoreUpdate, components []model.DurationComponent,
	) model.ScoreUpdate {
		switch typedUpdate := update.(type) {
		case model.Note:
			typedUpdate.Duration.Components = components
			return typedUpdate
		case model.Rest:
			typedUpdate.Duration.Components = components
			return typedUpdate
		default:
			return update
		}
	}

	durationComponents := getDurationComponents(update)
	if len(durationComponents) == 0 {
		return update
	}

	// We do preprocessing to obtain some necessary information
	var lastNoteLength model.NoteLength
	var lastNoteLengthIndex int
	noteLengthCount := 0

	for i, duration := range durationComponents {
		if reflect.TypeOf(duration) == noteLengthType {
			lastNoteLength = duration.(model.NoteLength)
			lastNoteLengthIndex = i
			noteLengthCount++
		}
	}

	if noteLengthCount == 1 &&
		lastNoteLength.Denominator == processor.currentDuration {
		// Only in this specific case do we have a repeated duration to remove
		// If noteLengthCount > 1, then we are dealing with ties
		durationComponents = append(
			durationComponents[:lastNoteLengthIndex],
			durationComponents[lastNoteLengthIndex + 1:]...
		)

		if len(durationComponents) == 0 {
			durationComponents = nil
		}
	}

	processor.currentDuration = lastNoteLength.Denominator
	update = setDurationComponents(update, durationComponents)
	return update
}
```

daveyarwood (approved):
This is fantastic. Nice work! :100: 

Just a naming suggestion: I think in compiler parlance, you could describe this sort of post-processing as "optimization". I think I like the idea of calling it `optimize` instead of `postprocess`, as it's more descriptive of what it's doing. Usually when you talk about optimization passes in a compiler, the optimizations are to improve the performance of the generated code, but I think making the generated code more idiomatic could also be thought of an optimization. What do you think?

Scowluga:
I agree optimizer is a better term, I didn't quite make that connection. I will rename and add an additional step to map midi note numbers to pitch identifiers.

Scowluga:
I have added a few commits. 

## Midi Note Number Translation (Commit 1, 4)
I've extracted the `NoteLetterIntervals: map[NoteLetter]int32` in `pitch.go` and used it to define the `func (mnn MidiNoteNumber) ToNoteAndOctave() (LetterAndAccidentals, int32)` mapping method. 

In commit 4 I added documentation and renamed this to a better name `ToLetterAndAccidentalsAndOctave`. 

I didn't add any tests for this because I think it is tested with the MusicXML `TestPercussion` test which calls post processor and tests it. 

## Renaming (Commit 2, 3)
I renamed post processing to optimizer. This is a much better name. 

Let me know and I'll merge both PRs.

client/model/pitch.go
```diff
@@ -202,3 +202,30 @@ func (mnn MidiNoteNumber) CalculateMidiNote(
...
+
+	// Find the closest letter than is above the current remainder
+	noteLetter := C
+	for letter, interval := range NoteLetterIntervals {
+		if interval <= remainder && interval > NoteLetterIntervals[noteLetter] {
+			noteLetter = letter
+		}
+	}
+
+	// Add necessary accidentals
```
daveyarwood, line 223: I think there might be a subtle bug here because of the way this could interact with global key signatures. Consider the following score:

```
(key-sig! '(e major))

piano:
  ...

percussion:
  c d e f g
```

If the percussion MIDI notes happen to be an F, C, G, or D (natural), and we don't emit a natural accidental, they will effectively end up becoming F#, C#, G# and D# because of the key signature.

On the other hand, it is weird that we apply global key signatures to percussion to begin with. So another way to address the same problem would be to special-case percussion parts so that key signatures don't apply.

I'm not sure which approach is better. What do you think?

Scowluga: This is a great point, I never considered this. So it looks like this problem doesn't affect MusicXML import as we never generate global key signatures and unpitched percussion parts can't have part-level key signatures (in MuseScore you physically cannot apply one). 

That being said, the translation method is currently in `pitch.go` so another developer could naively use it and fall victim to this mistake. You also pointed out another issue which is from https://github.com/daveyarwood/music-theory/issues/8. Essentially, when converting from `MidiNoteNumber` to `LetterAndAccidentals` we should have some knowledge of the key signature to apply the correct accidentals. For this mapping, it only applies to unpitched percussion which do not care about key signature. But for a more general `MidiNoteNumber` to `LetterAndAccidentals` translation we should be aware of this. 

For this reason, we've agreed the best solution is to move the code block from `model/pitch.go` to `importer/util.go` for now.

PR #372, opened 2021-07-12
MusicXML Importer - Reorganization

Scowluga:
I took a look at general code organization within Alda, and made some changes in the importer to better follow convention: 
1. Even ridiculously simple comments should be added for important public methods. e.x. for `ParseFile` or `ScanFile`. This is for organization, but also to auto-generate docs. I've added comments to `ImportMusicXML`.
2. The public method should be at the bottom of the file. I've adjusted `ImportMusicXML` to the bottom. 
3. Multiplexed methods like importer handlers, scanner `parseOctaveSet` / `parseString`, etc. should be methods on the struct itself. So like `scanner.parseString` instead of `parseString(... s *scanner)`. I didn't change this one within the MusicXML importer because it's just way too complex to put the handlers in the same file. 

Btw the diff on this PR is nasty, but I only moved code blocks around and added 2 lines of comments. No code changes, don't be scared 😄

### DIFF
--- a/client/interop/musicxml/importer/importer.go
+++ b/client/interop/musicxml/importer/importer.go
@@ -1,49 +1,14 @@
 package importer
 
 import (
+	"alda.io/client/color"
+	"alda.io/client/help"
+	"github.com/beevik/etree"
 	"io"
 	"sort"
 
-	"alda.io/client/color"
-	"alda.io/client/help"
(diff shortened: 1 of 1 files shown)


daveyarwood (approved):
That was easy to review. Looks good!

PR #373, opened 2021-07-14
Decompiler - Setup 

Scowluga:
#### The need for an Alda decompiler is twofold: 
1. With the start of interoperability, we are now able to import MusicXML into Alda IR (`ScoreUpdate`). Although we can do lots of stuff from this representation like output to JSON or play, we can't actually obtain Alda code itself from it. The decompiler will serve to decompile `ScoreUpdate` structs to Alda code so users can start writing Alda from imported music. 
2. Many great languages come with great formatters (ex `gofmt`). Alda may eventually aim to support built-in formatting. This is most easily done by scanning to `parser.Token` then formatting back to code. This PR will set up this formatter in the most basic form. 

#### This PR sets up the decompiler by: 
1. Setting up packages and public methods `code_generator.Generate` and `code_formatter.Format`. 
2. Updating the MusicXML importer to use these method (and remove the bootstrapped play implementation). 
3. Refactor `parser.Token` parameters to be public so they can be used outside of the `parser`. 

#### Alda compiler AST refactor 
This PR will wait for a refactor of the Alda parser to include a step to transform into an AST instead of a flat `[]parser.Token`. Then the generator will generate directly to this, and the formatter will get this as input.


client/parser/scanner.go
```diff
@@ -97,10 +97,10 @@ const (
...
 // A Token is a result of lexical analysis done by the scanner.
 type Token struct {
-	sourceContext model.AldaSourceContext
-	tokenType     TokenType
-	text          string
-	literal       interface{}
+	SourceContext model.AldaSourceContext
+	TokenType     TokenType
+	Text          string
+	Literal       interface{}
```
daveyarwood, line 103: I think once we do the refactor to have the parser produce an AST, we might not need these to be public. I'm thinking that tokens would remain an internal implementation detail of the parser, which turns text -> tokens -> AST. I don't think anything outside of the parser will be concerned with tokens, so we should keep them private if possible.

client/parser/scanner.go
```diff
@@ -664,7 +664,7 @@ func (s *scanner) scanToken() error {
 		// skip whitespace
 		return nil
 	case '#':
-		s.skipComment()
+		s.skipComment() // TODO: scan comments, only drop at the parsing level
```
daveyarwood, line 667: I thought a little bit about whether the AST should include comments. I think that it _has to_ include comments, in order to support code formatting, otherwise we would lose your comments when we reformat your code.

It seems like it might be useful if comments could be associated with other AST nodes, to potentially unlock annotation-like features in the future, so I'm thinking that maybe there could be a CommentedNode node, which combines an ASTNode with a comment string and a description of where, textually, the comment should go (is it on its own line? is it an inline comment at the end of the line?)

daveyarwood (commented):
Looks good so far. I think parts of this may change as we tackle the refactor to have the parser produce an AST, so let's keep this open for now.

client/parser/scanner.go
Scowluga: Right so if we use an AST representation, then the model/code generator packages where the code gen logic lies won't have to touch the actual token representation. 

That is likely a change we will make. Good point.

client/parser/scanner.go
Scowluga: Hmm so the question here is how do we determine the formatting for a comment (newline vs sameline)? 

One way is to have a commented node. If there is a commented node, it will go immediately after. Otherwise if it is a normal comment it's just on a newline. 

Another thing I was thinking about is dealing with custom formatting. Perhaps we should extend the scanner to capture extra whitespace. Then in our formatter there will be instances (such as around comments) where we leave the user-defined whitespace. I'm not sure this is necessary though, and I would be ok with just the concept of a commented node.

Scowluga:

PR #374, opened 2021-07-14
Decompiler - Code Generator

Scowluga:
This PR will introduce the Alda code generator as part of the complete decompiler. 

The code generator exists to process Alda IR that has no corresponding Alda code file. Currently, the only use case for this is the MusicXML importer.

### DIFF
--- a/client/code_generator/generator.go
+++ b/client/code_generator/generator.go
@@ -54,12 +54,22 @@ func (g *generator) generateBarline(_ model.Barline) {
 }
 
 func (g *generator) generateChord(chord model.Chord) {
-	for index, update := range chord.Events {
-		g.generateTokens([]model.ScoreUpdate{update})
+	hasNoteOrRestBeenEncountered := false
 
-		if index < len(chord.Events) - 1 {
+	for _, update := range chord.Events {
+		updateType := reflect.TypeOf(update)
+		isNoteOrRest := updateType == reflect.TypeOf(model.Note{}) ||
+			updateType == reflect.TypeOf(model.Rest{})
+
+		if isNoteOrRest && hasNoteOrRestBeenEncountered {
 			g.addToken(parser.Separator, "/")
 		}
+
+		if isNoteOrRest {
+			hasNoteOrRestBeenEncountered = true
+		}
+
+		g.generateTokens([]model.ScoreUpdate{update})
 	}
 }
 
@@ -67,6 +77,7 @@ func (g *generator) generateCram(cram model.Cram) {
 	g.addToken(parser.CramOpen, "{")
 	g.generateTokens(cram.Events)
 	g.addToken(parser.CramClose, "}")
+	g.generateDuration(cram.Duration)
 }
 
 func (g *generator) generateEventSequence(es model.EventSequence) {
@@ -117,19 +128,32 @@ func (g *generator) generatePitch(pitch model.PitchIdentifier) {
 }
 
 func (g *generator) generateDuration(duration model.Duration) {
+	hasEncounteredNonBarline := false
+	encounterBarline := func() {
+		if hasEncounteredNonBarline {
+			g.addToken(parser.Tie, "~")
+		}
+		hasEncounteredNonBarline = true
+	}
+
 	for _, component := range duration.Components {
 		switch value := component.(type) {
 		case model.Barline:
 			g.generateBarline(value)
 		case model.NoteLength:
-			noteLength := fmt.Sprintf("%f", value.Denominator) +
+			encounterBarline()
+
+			noteLength := formatFloat(value.Denominator) +
 				strings.Repeat(".", int(value.Dots))
 			g.addToken(parser.NoteLength, noteLength)
 		case model.NoteLengthMs:
-			noteLengthMs := fmt.Sprintf("%f", value.Quantity) + "ms"
+			encounterBarline()
+
+			noteLengthMs := formatFloat(value.Quantity) + "ms"
 			g.addToken(parser.NoteLengthMs, noteLengthMs)
-		case model.NoteLengthBeats:
+		default:
 			// NoteLengthBeats is only generated from Lisp
+			// Duration can only be nested... TODO
 			g.logError(component)
 		}
 	}
--- a/client/code_generator/generator.go
+++ b/client/code_generator/generator.go
@@ -1,10 +1,289 @@
 package code_generator
 
 import (
+	"alda.io/client/color"
+	log "alda.io/client/logging"
 	"alda.io/client/model"
 	"alda.io/client/parser"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
 )
 
+type generator struct {
+	tokens []parser.Token
+}
+
+func (g *generator) addToken(tokenType parser.TokenType, text string) {
+	g.tokens = append(g.tokens, parser.Token{TokenType: tokenType, Text: text})
+}
+
+func (g *generator) generateAttributeUpdate(au model.AttributeUpdate) {
+	switch partUpdate := au.PartUpdate.(type) {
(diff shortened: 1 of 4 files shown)


client/code_generator/generator.go
```diff
@@ -1,10 +1,295 @@
...
+	)
+}
+
+func (g *generator) generateVoiceGroupEndMarker(_ model.VoiceGroupEndMarker) {
+	g.addToken(parser.VoiceMarker, "V0:")
+}
+
+func (g *generator) generateTokens(scoreUpdates []model.ScoreUpdate) {
+	for _, update := range scoreUpdates {
+		switch value := update.(type) {
```
daveyarwood, line 241: I was thinking that this could be a part of the ScoreUpdate interface, something like this:

```go
type ScoreUpdate interface {
	...

	// ASTNode returns an AST node representation of the ScoreUpdate.
	ASTNode() ast.Node
}

func (vm model.VoiceMarker) ASTNode() ast.Node {
	return ast.Node{...}
}

func (g *generator) generateAST(scoreUpdates []model.ScoreUpdate) {
	for _, update := range scoreUpdates {
		g.addASTNode(update.ASTNode())
	}
}
```

I think this would be nice because the compiler would force us to implement the ASTNode method for every new kind of ScoreUpdate that we introduce in the future.

(This is all a bit speculative, of course, because we haven't done the refactor to have the parser produce an AST yet.)

client/code_generator/generator.go
```diff
@@ -1,10 +1,295 @@
...
+			case model.Sharp:
+				g.addToken(parser.Sharp, "+")
+			}
+		}
+	default:
+		// MidiNoteNumber has no direct representation in Alda syntax, and is
+		// only ever obtained from Lisp
+		// The MusicXML parser uses MidiNoteNumber pitches for percussion notes
+		// but translates this to LetterAndAccidentals in the optimization phase
+		g.logError(pitch)
```
daveyarwood, line 129: I think it might be reasonable in this case for the code generator to generate a Lisp AST node like `(note (pitch (midi-note 42)))`

daveyarwood (commented):
This is definitely a step in the right direction, even though we'll end up adjusting it following the parser refactor to produce an AST. :+1:

client/code_generator/generator.go
Scowluga: Right so in my initial research, this wasn't a possibility. This is because Golang prohibits cyclic module dependencies. So `parser` already needs `model` obviously, and this would require `model` need `parser`. 

If we can find some way to circumvent this, it'll allow us to add the generation directly to the `ScoreUpdate` interface. If we directly use Golang's `ast.Node` that would work. If we want to define our own AST node, we would have to think of some code organization that let's us introduce a dependency within `model` that doesn't violate Go's rules. 

Either way I mostly opened these PRs as an initial guess of my ideas. It is very easy to refactor the logic here into interface implementations within `model`.

client/code_generator/generator.go
Scowluga: So far I've chosen to not deal with any Lisp within the code generator, simply because the MusicXML importer will never generate it. In fact there's a few more things I've chosen to ignore as well. But if we're going to go towards having code generation be a `ScoreUpdate` interface, now we should consider how to approach all the methods I left unhandled. 

The problem here is I'm not really sure how to handle most of those `ScoreUpdate` elements. Most of them have no equivalent syntactic representation in Alda code, and can only be parsed to `ScoreUpdate` via Lisp and then possibly evaluating that Lisp. 

## Dealing with `ScoreUpdate`s
So one option is to somehow just generate Lisp from all of these `ScoreUpdate`s that don't have corresponding syntax. This would be hard, and I don't understand enough about Alda to even be sure if this works. **My intuition is it's possible to handle code generation for any arbitrary Alda `ScoreUpdate`**. 

For example, what if there's a note with a nice letter and accidentals pitch, a nice note length ms duration, but also a note length beats duration? This is valid Alda internal representation. But the only way to generate this would be to generate the entire `ScoreUpdate` as Lisp. This will cause problems for sure. 

So `ScoreUpdate` is fundamentally a superset of Alda source code. And if we try to embed code generation as a `ScoreUpdate` interface, there will be times we have to do weird workarounds or return error. I'm not as familiar with Alda as I probably should be to talk about this, but this is the understanding I have so far. 

## Keeping things the same
Then the other option is to simply leave it as is. We don't spin off code generation as a `ScoreUpdate` interface. We keep all of the logic in one place so we can easily see what is handled and what isn't. I think there's an argument for this. 

Code generation will only be used when importing from external sources. And when someone is writing that importer, they will be putting in a lot of work, which involves testing if it can be completely decompiled into source code. In this process they will have to look at the code generator to make sure everything they build can be generated. Then support can be added as necessary. 

Another option is to really make this distinction clear. We can put `code_generator` in the `interop` folder. Then `code_formatter` in the base `client` folder. This is because formatting can be done on Alda in general, and is closer to the language itself. But generation is only on imported representations which will happen in the `interop` folder.

client/code_generator/generator.go
daveyarwood: I had that same thought. I was thinking that we could put our AST related code in a separate module from parser and model. I wasn't aware that there even was a Go AST package! :joy:  It looks like [ast.Node](https://pkg.go.dev/go/ast#Node) is not necessarily what we want, it's just an interface with a couple of functions related to the node's position in the text.

I'm still up for taking a crack at that refactor myself, BTW. Just a matter of finding some time for it!

client/code_generator/generator.go
daveyarwood: I had a glance over all of the various things that implement the ScoreUpdate interface, and I believe it should be possible for us to add `AST()` as an interface method. There are a couple of interesting ones, like:

* AttributeUpdate
  * Has a PartUpdate, so we could also add `AST()` to the PartUpdate interface and have the compiler force us to define how each PartUpdate would be represented e.g. in the inline Lisp. I would view this as a positive thing, because it would encourage us to make sure that all kinds of part updates are supported in Lisp.
* GlobalAttributeUpdate
  * Also has a PartUpdate, but the only difference is that we need to generate e.g. `volume!` instead of `volume`. I guess one way we could do this is to add another method to the PartUpdate interface to explicitly define how the _global_ version of that part update would be represented as an AST. That feels kind of gross, but I think I'd be willing to accept the grossness as a trade-off for the benefits of this approach where the compiler helps us make sure that everything has an AST representation.
* LispNil, LispList
  * These are the only two things in lisp.go that implement ScoreUpdate, and they would be trivial to represent in the AST, because of the one-to-one relationship of Lisp code to the AST. When we implement the AST, I think we'll probably end up just having a corresponding AST node type for each kind of value possible in the Lisp, e.g. an AST LispString that corresponds to model.LispString, and the same thing for LispNumber, LispList, etc. etc.
* Note, Rest
  * These will be a bit tricky, as you mentioned, but not _too_ tricky, I don't think.
  * We can special-case notes where the pitch is represented as a MIDI note number and choose to return the note in Lisp form.
  * As for the thing about durations: I don't think it's possible for a Note to have a duration expressed in both beats and milliseconds simultaneously. A Note has a Duration, which has a list of DurationComponents. DurationComponent is an interface with specific implementations including NoteLength, NoteLengthBeats, and NoteLengthMs. We could add an `AST()` function to the DurationComponent interface as well, so that the compiler ensures that we always define how to generate the AST for these things. If there are multiple duration components, then we tie them together with a tie (`~`) node in between each component.

client/code_generator/generator.go
Scowluga: Haha ok. Sounds about right. I thought Go's build-in AST would not be what we wanted. A new package seems correct.

Scowluga:

PR #376, opened 2021-07-14
MusicXML Importer - Test Case Reformatting

Scowluga:
I've reformatted the MusicXML importer test case file `importer_test.go`. 

1. I realize for some reason I implemented an `executeImporterTestCases` method taking in a varargs `...importerTestCase` parameter but I never used this feature. I've changed to use this for repeated tests like for repeat1, repeat2, repeat3. 
2. I indented all Alda code blocks. This just fits better, especially with the use of varags above. 

This is a simple change. As with before, feel free to just merge PRs like this.

### DIFF
--- a/client/interop/musicxml/importer/importer_test.go
+++ b/client/interop/musicxml/importer/importer_test.go
@@ -1,64 +1,60 @@
 package importer
 
 import (
+	"alda.io/client/model"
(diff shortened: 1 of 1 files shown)


daveyarwood (approved):
:ok_hand:

PR #393, opened 2021-08-18
Added Mr Tambourine Man as if played on guitar example

t0mk:
Hey, I've just dicovered alda a couple of hours ago and tried to put together Mr. Tambourine Man as I'd play it on guitar.

I think the code is a bit clumsy because I didn't really figure how make variable of a chord, and then use the var with different lengths.

Thanks for the work on alda, it took me really short time to start. You don't have to merge this in if you don't want, but you can give me feedback how to maybe work with chords more efficiently in alda.

### DIFF
--- a/examples/mr_tambourine_man.alda
+++ b/examples/mr_tambourine_man.alda
@@ -1,0 +1,28 @@
+# This is Mr. Tambourine Man as I would play it on a guitar with barre chords,
+# with 8 strums (up-down is 2 strums).
+
+# Chords with different durations, e.g "gc" is a G barre chord on guitar
+# with duration of 1 bar. e.g. "gc3" is a G chord with duration of 3 beats
+# (3 eights).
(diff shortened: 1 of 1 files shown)


daveyarwood:
Nice work! :100: 

I will pass on including this on the examples, only in the interest of keeping the set of examples small and focused on specific Alda features. But I did play around with this a little and came up with a way to use a variable avoid repeating the notes in each chord:

```alda
gmaj = o2 g>/d/g/b>/d/g~
gc = (set-note-length 8) gmaj
gc2 = (set-note-length 4) gmaj
gc3 = (set-note-length "4.") gmaj
```

I wish there were a nice, concise way to take a chord (or perhaps a variable?) and repeat it with a different duration, but that's a problem I've never been able to solve with the design of Alda. If anyone reading this has thoughts, it might be a good thing to brainstorm about!

t0mk:
@daveyarwood thanks for the tips!

Issue #321, opened 2019-03-07
Validate MIDI note range


daveyarwood:
Currently, attempting to play a note that is out of the MIDI note range (0-127) results in a confusing error message:

```
$ alda play -c 'piano: o10 c'
[27713] Parsing/evaluating...
[27713] ERROR data1 out of range: 132
```

We should catch this and throw our own error with a clearer message.

daveyarwood:
I just tried this with Alda 2, and interestingly, playback seems to just silently fail once it gets up to the too-high note. No error message is printed either in the client or in the player logs.

So the disposition is still the same: we should add validation that the MIDI note isn't < 0 or > 127, and I think we should do that validation at score evaluation time instead of playback-time, so that we can display a helpful message earlier.

I think the place to add the validation now is [right here](https://github.com/alda-lang/alda/blob/48ea575aad588c1288116b8289a865af9231583b/client/model/note.go#L120-L122) after we calculate the MIDI note based on the octave, note letter and accidentals.

KartikYZ:
I have added the validation at the specified place and submitted a PR in response to this issue. Also ran the addition against the test suite.

daveyarwood:
Fixed in https://github.com/alda-lang/alda/pull/461. Will keep this issue open until I have a chance to release it.

daveyarwood:
Fixed in Alda 2.2.4.

PR #409, opened 2021-12-31
#321 Validate MIDI note range

KartikYZ:
Added validation for midi note range in note.go. We raise a UserFacingError if midi note is > 0 or < 127 and log the event.

### DIFF
--- a/client/model/note.go
+++ b/client/model/note.go
@@ -127,6 +127,10 @@ func addNoteOrRest(score *Score, noteOrRest ScoreUpdate) error {
 				)
 
 				if midiNote < 0 || midiNote > 127 {
+					log.Debug().
+						Int("midiNote", int(midiNote)).
+						Msg("Midi note out of range.")
+
 					return help.UserFacingErrorf("Midi note out of the 0-127 range. Input note: %d", midiNote)
 				}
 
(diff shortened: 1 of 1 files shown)


client/model/note.go
```diff
@@ -125,6 +126,14 @@ func addNoteOrRest(score *Score, noteOrRest ScoreUpdate) error {
 					part.Octave, part.KeySignature, part.Transposition,
 				)
 
+				if midiNote < 0 || midiNote > 127 {
+					log.Debug().
+						Int("midiNote", int(midiNote)).
+						Msg("Midi note out of range.")
```
daveyarwood, line 132: This debug log has the same information as the user-facing error message, so I think we can remove it to reduce noise.

daveyarwood (commented):
This looks good, thanks for the PR! My two suggestions would be to remove the debug log message, and to add tests. I saw you were asking about how to approach writing tests on Slack, so I'll answer there.

daveyarwood:
Fixes #321.

client/model/note.go
KartikYZ: Thanks for the feedback, I've checked the Slack. Working on it!

daveyarwood:
Superceded by #461

Issue #404, opened 2021-10-22
REPL :stop command fails for overly long segments


elyisgreat:
# 🐞 Bug report 🐞

## Description

The REPL `:stop` command simply fails to function when an overly long segment is played. More specifically, the `:stop` command simply fails after the first 2 seconds for certain sufficiently long segments (usually around 45 seconds or longer). Quitting the repl and issuing an `alda stop` or `alda shutdown` command seems to be the only way to stop the playback.

## Steps to Reproduce
1. Run the alda repl
2. Enter `piano: c*100` into the repl
3. Wait for a couple of seconds. After the first four or five notes, issue a `:stop` command.

## Expected Behavior
The segment stops playing.

## Actual Behavior
The segment fails to stop playing.

## Environment

Windows 10, PowerShell, alda 2.0.6

ksiyuan:
Hello, I want to solve this issue but have stuck for a while. This bug doesn't always happen on my computer, but it happens more frequently as the fragment gets longer. I might be totally wrong but is it possible that this have something to do with race condition? Any thought would be appreciated.

daveyarwood:
I have only a fuzzy understanding of how this might be happening. The first thing I would check is whether the "stop" message is making it to the player process. If you run `alda -v2 stop` (with verbosity level 2 or higher), you can see the INFO logs that show the "stop" messages being sent from the client to each player process. So that's one thing to check -- is the player process that's currently playing included in the list of player processes that the client is sending the "stop" message to?

Another thing to check is whether the player process is actually receiving the "stop" message and processing it right away. You can run a player process in the foreground in "very verbose" mode by running `alda-player -vv run`. Note the three-letter ID at the beginning of each log message. You can then use `alda play -i xxx` (where `xxx` is the ID of your player process) to send "play" and "stop" messages.

I was playing around with this just now, and here is some the of output of the player process while I ran `alda play -i iqt -c "piano: c*100"` and then `alda stop -i iqt` in another terminal:

```
iqt INFO  2022-04-13 20:42:01 StateManager.cleanUpStaleStateFiles:72 - Cleaning up stale files in /home/dave/.cache/alda/state/players...
iqt INFO  2022-04-13 20:42:01 StateManager.cleanUpStaleStateFiles:72 - Cleaning up stale files in /home/dave/.cache/alda/state/repl-servers...
iqt INFO  2022-04-13 20:42:01 Main.run:77 - Starting receiver, listening on port 38447...
[...]

PR #418, opened 2022-04-16
#404 REPL :stop command fails for overly long segments

ksiyuan:
I changed the function hasPlayer so that it checks whether server.player is not an empty struct instead of only checking if server.player.Port is equal to 0. After that it seems the issue never comes out on my computer but I don't know the exact reason. 

Besides, to give user a choice to shut down all player process, I added another repl command called stopall which shut down all player process while didn't change the original stop command so that the user can choose whether to shut down the current player or simply shut down everything if something wrong happens.


daveyarwood:
Thanks for the PR, but I think I would prefer another approach where we make sure that `server.player` doesn't end up being set to something unexpected. The improvements in this PR are nice, but they feel more like workarounds than a solution to whatever the problem is.

Assuming [my theory here](https://github.com/alda-lang/alda/issues/404#issuecomment-1100743439) is correct: If the current player is valid but there happens to be an error reading its state file and the port is unexpectedly set to 0, then we should leave `server.player` alone and continue to interact with that same player process. I have a hunch that that might fix it :crossed_fingers:

I think there are two ways we could do this (and maybe it makes sense to do both?)

1. In the places where we set `server.player`, we should check that the port isn't 0 and avoid setting it if that's the case.
2. In [ReadPlayerStates](https://github.com/alda-lang/alda/blob/2ffdb23f12486398310bf871e403038bae68be10/client/system/process_management.go#L165-L199), we probably shouldn't include any player states where there was a read error in the results.

ksiyuan:
I tried both way. 
1. Check if the new player/updated state has port = 0 and leave the current server.player unchanged if that's the case

2. In [ReadPlayerStates](https://github.com/alda-lang/alda/blob/2ffdb23f12486398310bf871e403038bae68be10/client/system/process_management.go#L165-L199), simply return `nil` as well as `readError` when there was a read error in the results.

Not sure if that solves all the problems, but it seems to work well on my computer.

daveyarwood:
Thanks for looking into this! :raised_hands: 

I'm going to make some small adjustments after merging, and get this into a release soon so that we can collectively confirm whether the problem goes away.

PR #422, opened 2022-10-02
Update comments.md

yuvihere23:


daveyarwood:
Hi! Thanks for the contribution, but I would prefer that we keep that word emphasized.

PR #457, opened 2022-10-10
update the shell file

ragu-1997:
shell module(label: low-hanging-fruit)

### DIFF
--- a/scripts/install-fluidr3.sh
+++ b/scripts/install-fluidr3.sh
@@ -18,3 +18,4 @@ mkdir -p ~/.gervill
 cp fluid-r3.sf2 ~/.gervill/soundbank-emg.sf2
 echo "Installed ${FLUID_NAME}"
 cd .. && rm -rf ${FLUID_WORK_FOLDER}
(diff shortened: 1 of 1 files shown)


ragu-1997:
Hi Alda,

Can you give me reason for spam because of my learning perspective.
Possible to remove the spam Flag because I want contribute more am a
fresher .

Happy learning 🎉

ragu-1997:
Thanks for your time Alda now am understand . can you remove spam Flag
because I want to participate.sorry for that

ragu-1997:
> * Adding `exit 0` to the end of a shell script does not change the behavior of the shell script at all. Shell scripts always exit 0 by default if they make it to the end without exiting with another status
> * Adding `exit 0` to the end of a shell script is not idiomatic Bash.
> * Adding `exit 0` to the end of the script does not provide any additional value over what the script is already doing.
> * This PR does not address any of the issues that I have labeled `hacktoberfest`.

Hey dav,
It's not kind of spam just a package please consider remove the spam tag

PR #461, opened 2022-11-14
Fix issue #321 - Validate MIDI Note range

kylewilk567:
Added an error when MIDI is out of the range 0-127 and added a test case checking the error is thrown.


daveyarwood (approved):
Looks great. Thanks for the contribution!

Issue #421, opened 2022-07-20
Alda REPL "skips" rests at the end of a line


daveyarwood:
# 🐞 Bug report 🐞

## Description

@litan on [Alda Slack](https://slack.alda.io) observed that if you run the following command twice in succession:

```bash
alda repl \
  --client \
  --port 43009 \
  --message '{"op": "eval-and-play", "code": "percussion: o2 (tempo 240) c4 r4 c4 r4 | c4 r4 c4 r4 | c4 r4 c4 r4 | c4 r4 d4 r4"}'
```

It appends the two together as expected, however, the `r4` at the end gets stripped off, so the `c4` at the beginning of the second command ends up happening a quarter note too soon.

I was also able to reproduce this in an interactive Alda REPL session.

## Steps to Reproduce

1. Run `alda repl` to start a REPL session.
2. Enter `percussion: o2 (tempo 240) c4 r4 c4 r4 | c4 r4 c4 r4 | c4 r4 c4 r4 | c4 r4 d4 r4` and press Enter.
3. Before that line finishes playing, press Up arrow, Enter to enter the same line again.

## Expected Behavior

The notes you hear are all spaced apart by quarter note rests.

## Actual Behavior

It sounds like the `r4` quarter note rest at the end of the first line is omitted, and there is a "hiccup" where the second line starts a quarter note too early.

## Environment

**Operating system and version:** Ubuntu 20.04

**Alda version:** 2.2.3

Samyra312007:
Respected Maintainers, If this issue is still open then i would like to work on this issue.

daveyarwood:
@Samyra312007 Please feel free!

Samyra312007:
I have opened PR #522 
Kindly review it and let me know if i am wrong somewhere with my implementation.

daveyarwood:
Fixed in Alda 2.4.4, which will be released shortly. Thanks @Samyra312007! 🙌

PR #462, opened 2022-11-20
Add check for extra rest duration

kylewilk567:
Patch to fix issue #421 . Checks the ending duration of what is sent to the client matches the PartOffset in the Score object. If there is a mismatch (as in the case with a trailing rest), extra duration is appended to the last note.

### DIFF
--- a/client/transmitter/osc.go
+++ b/client/transmitter/osc.go
@@ -311,7 +311,7 @@ func (oe OSCTransmitter) ScoreToOSCBundle(
 	// of the score.
 	scoreLength := 0.0
 
-	for _, event := range events {
+	for index, event := range events {
 		eventOffset := event.EventOffset()
 
 		// Filter out events before the `--from` time marking / marker, when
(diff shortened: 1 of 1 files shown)


client/transmitter/osc.go
```diff
@@ -327,6 +327,21 @@ func (oe OSCTransmitter) ScoreToOSCBundle(
...
+			// Additional rests at the end of a code sequence create a timing mismatch
+			// between Part.CurrentOffset and this note event's Offset + duration
+			// Part.CurrentOffset should equal Note Offset + Note Duration, so this statement
+			// Increments Note Duration so that they become equal
+			restOffset := int32(0)
+			if index == len(events)-1 {
+				lastEvent := event
+				// Last event offset + duration must equal score.duration
+				currNoteEnd := int32(math.Round(lastEvent.Duration)) + int32(lastEvent.EventOffset())
+				partOffset := int32(score.CurrentParts[len(score.CurrentParts)-1].CurrentOffset)
```
daveyarwood, line 341: I think selecting the last Part is potentially error-prone, as it might not be the Part that the note belongs to. I think `lastEvent.Part` is probably what we want here.

daveyarwood:
Thanks for the PR, Kyle! It looks good overall, apart from the thing I mentioned above.

I'll do some manual testing soon when I have time, and merge it if everything looks and sounds good.

daveyarwood:
This doesn't appear to fix the issue in #421, unfortunately :(

Here's what I did to test:

* Run `alda repl -s -p 12345` to start a REPL server on port 12345.
* Check out this branch.
* Run the following command twice in rapid succession:

```
client/bin/run repl \
  --client \
  --port 43009 \
  --message '{"op": "eval-and-play", "code": "percussion: o2 (tempo 240) c4 r4 c4 r4 | c4 r4 c4 r4 | c4 r4 c4 r4 | c4 r4 d4 r4"}'
```

The "hiccup" is still there where the final rest is skipped.

kylewilk567:
I tested the code using the following commands in Windows Powershell and it works:
image

I wasn't able to replicate it using the procedure above since my Alda executable in my PCs PATH is the old version. Do you have any suggestions as to why our testing may have given different results?

daveyarwood:
@kylewilk567 Sorry for the late reply - I've been on vacation the last couple weeks.

This is interesting! I tried it the way you described, and I also heard the last rest being handled correctly. But I tried it again the way I described (running a REPL server in another terminal and using the CLI to send it messages), and the rest is still getting left out when I do it that way.

I wonder if there is some important difference in the way the messages are being sent to the REPL server when you do it via the CLI vs. doing it interactively in a REPL client session?

daveyarwood:
Closing due to inactivity.

PR #471, opened 2023-02-21
ASTNode Generation from []ScoreUpdate's

Scowluga:
# Background

This PR is the first step of the process, introducing `[]ScoreUpdate` -> `ASTNode` conversion via the following function in `parser/gen.go`:
```
func GenerateASTFromScoreUpdates(updates []model.ScoreUpdate) (ASTNode, error)
```

# Notes
### `gen.go`
The generation is mostly straightforward, with a few points to note:
1. As commented, `AldaSourceContext` is ignored in generation. I think this is a very fine choice.
2. Processing parts and voices is a little non-trivial. I tried to do it in the cleanest way, but essentially the generator must "reconstruct" an `ASTNode` because `ASTNode.Updates` "flattens" this information in `ScoreUpdate`s.

### `PartUpdate` to `LispListNode`
We only handle conversion for `PartUpdate`'s that are introduced by the MusicXML importer.
We also add `KeySignature.String()` so that we can generate the correct literal for the corresponding `LispStringNode` of a `KeySignatureSet`.

### `test_helper.go`
I've slightly refactored/generalized `parser`'s tests in `parseTestCase` in `test_helper.go`:
1. Now providing actual `ASTNode` or `[]ScoreUpdate` is optional. These are tested if provided. Otherwise, we just test that everything works and that the generated `ASTNode` is equal the parsed `ASTNode`.
2. We now always test applying `ScoreUpdate`s to a `Score` (with an opt out). I also changed a couple test cases so that they could be applied to a score. Let me know if we want it changed.
5. `examples_test.go` which tests the example files now uses `parseTestCase`.

### `e2e`
We've talked about a complete refactor of testing to spin-off an `e2e` testing module downstream of both `parser` and `model`. I think the changes we have now are "good enough" that this change is not quite worth it for now, given my timeline. I've left it as a TODO.

### DIFF
--- a/client/doc/notes/interaction-modes.adoc
+++ b/client/doc/notes/interaction-modes.adoc
@@ -76,7 +76,7 @@ development and live coding.
 * One use case is for each collaborator to work on a separate part.
 ** It might be useful for the REPL client session (or text editor plugin, etc.)
 to keep track of which instrument the collaborator used last, and explicitly
-prepend that part declaration (e.g. `bass:`) to the beginning of every message
+prepend that part declaration (e.g. `piano:`) to the beginning of every message
 to the server.
 *** In an interactive Alda REPL client session, we could actually make the part
 declaration an editable part of the message, so that the user could see that the
--- a/client/model/key.go
+++ b/client/model/key.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"math"
+	"strings"
 
 	"alda.io/client/json"
 )
@@ -44,6 +45,30 @@ const (
 //
 //   {B: [flat, flat], G: [sharp], E: [flat]}
 type KeySignature map[NoteLetter][]Accidental
+
+func (ks KeySignature) String() string {
+	laas := []string{}
+
+	for note, accidentals := range ks {
+		laa := strings.Builder{}
+		laa.WriteRune(rune(note + 'a'))
+
+		for _, acc := range accidentals {
+			switch acc {
+			case Flat:
+				laa.WriteString("-")
+			case Natural:
+				laa.WriteString("_")
+			case Sharp:
+				laa.WriteString("+")
+			}
+		}
+
+		laas = append(laas, laa.String())
+	}
+
+	return strings.Join(laas, " ")
+}
 
 // JSON implements RepresentableAsJSON.JSON.
 func (ks KeySignature) JSON() *json.Container {
--- a/client/parser/ast.go
+++ b/client/parser/ast.go
@@ -696,7 +696,7 @@ func (node ASTNode) Updates() ([]model.ScoreUpdate, error) {
 		}, nil
 
 	case PartNode:
-		if err := node.expectChildren(); err != nil {
+		if err := node.expectNChildren(2); err != nil {
 			return nil, err
 		}
 
@@ -734,23 +734,17 @@ func (node ASTNode) Updates() ([]model.ScoreUpdate, error) {
 			partDecl.Alias = partAlias.Literal.(string)
 		}
 
-		updates := []model.ScoreUpdate{partDecl}
-
-		if len(node.Children) > 1 {
-			events, err := node.Children[1].expectNodeType(EventSequenceNode)
-			if err != nil {
-				return nil, err
-			}
-
-			eventUpdates, err := concatChildUpdates(events)
-			if err != nil {
-				return nil, err
-			}
-
-			updates = append(updates, eventUpdates...)
+		events, err := node.Children[1].expectNodeType(EventSequenceNode)
+		if err != nil {
+			return nil, err
 		}
 
-		return updates, nil
+		eventUpdates, err := concatChildUpdates(events)
+		if err != nil {
+			return nil, err
+		}
+
+		return append([]model.ScoreUpdate{partDecl}, eventUpdates...), nil
 
 	case RepeatNode:
 		if err := node.expectNChildren(2); err != nil {
@@ -892,14 +886,6 @@ func (node ASTNode) Updates() ([]model.ScoreUpdate, error) {
 			},
 			voiceEventUpdates...,
 		), nil
-
-	case VoiceNumberNode:
-		return []model.ScoreUpdate{
-			model.VoiceMarker{
-				SourceContext: node.SourceContext,
-				VoiceNumber:   node.Literal.(int32),
-			},
-		}, nil
 	}
 
 	return nil, fmt.Errorf(
--- a/client/parser/examples_test.go
+++ b/client/parser/examples_test.go
@@ -2,20 +2,15 @@ package parser
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"testing"
 
-	"alda.io/client/model"
 	_ "alda.io/client/testing"
 )
 
-// TestExamples parses and compiles each of the example scores in the `examples`
-// directory.
-//
-// Strictly speaking, this is not testing just the parser. Perhaps these tests
-// should live in another package that consumes both the `parser` and `model`
-// packages, but I'm not sure yet which package that should be.
+// TestExamples tests each of the example scores in the `examples` directory.
 func TestExamples(t *testing.T) {
 	dir, err := os.Getwd()
 	if err != nil {
@@ -25,9 +20,12 @@ func TestExamples(t *testing.T) {
 
 	examplesDir := filepath.Join(filepath.Dir(filepath.Dir(dir)), "examples")
 
-	err = filepath.Walk(examplesDir,
+	filepath.Walk(examplesDir,
 		func(path string, info os.FileInfo, err error) error {
+			label := fmt.Sprintf("examples test for %s", path)
 			if err != nil {
+				t.Error(label)
+				t.Errorf("filepath walk error %s", err)
 				return err
 			}
 
@@ -35,22 +33,19 @@ func TestExamples(t *testing.T) {
 				return nil
 			}
 
-			fmt.Printf("● %s\n", path)
-
-			ast, err := ParseFile(path)
+			contents, err := ioutil.ReadFile(path)
 			if err != nil {
+				t.Error(label)
+				t.Error("issue reading file contents")
 				return err
 			}
 
-			scoreUpdates, err := ast.Updates()
-			if err != nil {
-				return err
-			}
+			executeParseTestCases(t, parseTestCase{
+				label: label,
+				given: string(contents),
+			})
 
-			score := model.NewScore()
-			return score.Update(scoreUpdates...)
-		})
-	if err != nil {
-		t.Errorf("%v\n", err)
-	}
+			return nil
+		},
+	)
 }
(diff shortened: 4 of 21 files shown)


client/parser/ast.go
```diff
@@ -736,20 +736,18 @@ func (node ASTNode) Updates() ([]model.ScoreUpdate, error) {
 
 		updates := []model.ScoreUpdate{partDecl}
 
-		if len(node.Children) > 1 {
```
Scowluga, line 739: @daveyarwood I have to check if I'm correct, but it doesn't seem like this branch is necessary. The parser always generates an `EventSequenceNode` here. This is in contrast to say `CramNode`, which has an optional second `EventSequenceNode` child.

client/parser/ast.go
```diff
@@ -892,14 +890,6 @@ func (node ASTNode) Updates() ([]model.ScoreUpdate, error) {
 			},
 			voiceEventUpdates...,
 		), nil
-
-	case VoiceNumberNode:
```
Scowluga, line 896: I have to check, but this also doesn't seem necessary. It's handled in the above `case VoiceNode` and never generated separately.

client/doc/notes/interaction-modes.adoc
```diff
@@ -76,7 +76,7 @@ development and live coding.
 * One use case is for each collaborator to work on a separate part.
 ** It might be useful for the REPL client session (or text editor plugin, etc.)
 to keep track of which instrument the collaborator used last, and explicitly
-prepend that part declaration (e.g. `bass:`) to the beginning of every message
```
Scowluga, line 79: @daveyarwood "bass" instrument no longer exists.

client/parser/parser.go
```diff
@@ -966,7 +966,7 @@ func (p *parser) voiceGroup() (ASTNode, error) {
 
 	voiceGroupNode := ASTNode{
 		Type:          VoiceGroupNode,
-		SourceContext: firstVoiceMarkerToken.sourceContext,
```
Scowluga, line 969: @daveyarwood small bug where we didn't suppress for this token.

client/parser/parser.go
```diff
@@ -10,13 +10,12 @@ import (
...
 	"alda.io/client/help"
 	log "alda.io/client/logging"
-	model "alda.io/client/model"
+	"alda.io/client/model"
 )
 
 type parser struct {
 	filename string
 	input    []Token
-	updates  []model.ScoreUpdate
```
Scowluga, line 19: @daveyarwood It doesn't look like `parser` needs to have this updates field so I've removed it.

client/parser/ast.go
daveyarwood: Seems reasonable, since I can't think of a case where we would only want to generate the AST for a VoiceNumberNode by itself :+1:

client/parser/examples_test.go
```diff
@@ -25,32 +20,32 @@ func TestExamples(t *testing.T) {
...
+				label: label,
+				given: string(contents),
+			})
 
-			score := model.NewScore()
-			return score.Update(scoreUpdates...)
-		})
-	if err != nil {
-		t.Errorf("%v\n", err)
-	}
```
daveyarwood, line 55: I think this part might be important, although I'm not 100% sure. I think without it, the `filepath.Walk(...)` expression could be returning an error, and we're just ignoring it.

daveyarwood: ...although, maybe `t.Error` and `t.Errorf` have the side effect of making the test case fail, so effectively it doesn't matter?

Just to be safe, could you test making one of the example files deliberately wrong and ensuring that the test case fails?

client/parser/parser.go
daveyarwood: Nice catch!

client/parser/parser.go
daveyarwood: Interesting, looks like you're right! Maybe that was a holdover from an earlier implementation. Seems safe to remove.

client/parser/test_helper.go
```diff
@@ -1,51 +1,83 @@
...
+// parseTestCase also tests Score Updating which tests the `model` module
+// Perhaps this should live in another package that consumes both `parser` and
+// `model`, but for now we keep it here
 type parseTestCase struct {
-	label  string
-	given  string
-	expect []model.ScoreUpdate
+	label            string
+	given            string
+	expect           []model.ScoreUpdate // optional
```
daveyarwood, line 18: We could consider calling this `expectUpdates` for extra clarity

client/parser/test_helper.go
```diff
@@ -1,51 +1,83 @@
...
 		}
+		if testCase.expectAST != nil {
+			diff := deep.Equal(testCase.expectAST, actualAST);
+			if diff != nil {
+				t.Error(testCase.label)
+				for _, diffItem := range diff {
+					t.Errorf("%v", diffItem)
+				}
+			}
+		}
```
daveyarwood, line 45: Nice :+1:

client/parser/test_helper.go
```diff
@@ -1,51 +1,83 @@
...
 
-		if diff := deep.Equal(testCase.expect, actualUpdates); diff != nil {
+		generatedAST, err := GenerateASTFromScoreUpdates(actualUpdates)
+		if err != nil {
+			t.Errorf("%v\n", err)
+			return
+		}
+		if diff := deep.Equal(actualAST, generatedAST); diff != nil {
```
daveyarwood, line 67: This is great! I think it might be good to add a comment to make our intent clear: we want to ensure that the "round trip" where we translate an AST into ScoreUpdates, and then back to an AST, results in an AST equivalent to the original one.

client/parser/variables_test.go
```diff
@@ -16,53 +17,37 @@ func variableReference(name string) model.VariableReference {
 	return model.VariableReference{VariableName: name}
 }
 
+func variableNameCheck(name string) parseTestCase {
+	return parseTestCase{
+		label: fmt.Sprintf("variable definition and reference: %s", name),
+		given: fmt.Sprintf("%[1]s = r \n%[1]s", name),
```
daveyarwood, line 23: TIL that you could do this. Mind blown.

client/parser/variables_test.go
```diff
@@ -16,53 +17,37 @@ func variableReference(name string) model.VariableReference {
...
-			expect: []model.ScoreUpdate{variableReference("underscores_are_great")},
-		},
+		variableNameCheck("aa"),
+		variableNameCheck("aaa"),
+		variableNameCheck("HI"),
+		variableNameCheck("celloPart2"),
+		variableNameCheck("xy42"),
+		variableNameCheck("my20cats"),
+		variableNameCheck("apple_cider"),
+		variableNameCheck("underscores_are_great"),
```
daveyarwood, line 41: Nice refactor :+1:

client/parser/gen.go
```diff
@@ -0,0 +1,391 @@
...
+
+	node.Children = append(node.Children, durationNode)
+	return node, nil
+}
+
+// mapIsolatedUpdate maps a single isolated model.ScoreUpdate to ASTNode.
+// Holistic updates that require "re-construction" are handled upstream:
+// 	1. model.PartDeclaration in GenerateASTFromScoreUpdates.
+// 	2. model.VoiceMarker, model.VoiceGroupEndMarker in mapUpdatesWithoutParts.
+func mapIsolatedUpdate(updates model.ScoreUpdate) (ASTNode, error) {
```
daveyarwood, line 60: `updates` is a confusing name for a single ScoreUpdate. Maybe this should be called `update`, and `update` below could just be `u`?

daveyarwood (commented):
I've left a handful of small comments, but this looks really good overall :raised_hands:

client/parser/gen.go
```diff
@@ -0,0 +1,391 @@
...
+		)
+
+	}
+}
+
+// mapUpdatesWithoutParts maps a group of updates to a group of ASTNode's.
+// Parts are ignored and handled upstream in GenerateASTFromScoreUpdates.
+// Voices are handled by identifying model.VoiceMarker then generating
+// subsequent nodes within an overarching VoiceGroupNode.
+func mapUpdatesWithoutParts(updates []model.ScoreUpdate) ([]ASTNode, error) {
```
Scowluga, line 279: note for self, makes more sense to call this `mapUpdatesWithinParts` or something like that

client/parser/gen.go
```diff
@@ -0,0 +1,391 @@
+package parser
+
+import (
+	"alda.io/client/model"
+	"fmt"
+)
+
+// appendDuration appends a DurationNode child for a non-empty model.Duration.
+func appendDuration(node ASTNode, duration model.Duration) (ASTNode, error) {
```
Scowluga, line 9: note to self, makes more sense to call this `withDuration` or something

client/parser/test_helper.go
```diff
@@ -1,51 +1,83 @@
...
+
+		// We suppress source context to facilitate deep diff comparison
+		actualAST, err := Parse(
+			testCase.label, testCase.given, SuppressSourceContext)
 		if err != nil {
 			t.Errorf("%v\n", err)
 			return
 		}
+		if testCase.expectAST != nil {
+			diff := deep.Equal(testCase.expectAST, actualAST);
```
Scowluga, line 38: note to self: extra semi colon

client/parser/variables_test.go
Scowluga: Haha I had to search it up too. Pretty neat though.

client/parser/gen.go
Scowluga: Refactored to follow parser naming format with "topLevel" and "innerEvents"

client/parser/examples_test.go
Scowluga: Good point. It looks like `t.Error` does make the test fail (without returning error). I think we `return err` within the walk more as a short-circuit (e.x. no point running test case if can't read file).

Resolving.

client/parser/test_helper.go
Scowluga: Good comment to add. Done.

client/parser/test_helper.go
Scowluga: Agreed. I actually made the change, then reverted it so that the PR would be more clean.

Now that you've done a first-pass, I'll do the renaming as it's own commit.

client/model/key.go
```diff
@@ -45,6 +46,30 @@ const (
 //   {B: [flat, flat], G: [sharp], E: [flat]}
 type KeySignature map[NoteLetter][]Accidental
 
+func (ks KeySignature) String() string {
```
Scowluga, line 49: @daveyarwood I added this pretty hastily. Let me know if something like it exists / it can be done more easily / we rename the function.

daveyarwood: This looks reasonable to me, and I don't think we have anything like it yet :+1:

daveyarwood (approved):
Looks good :+1:

PR #474, opened 2023-03-26
Alda Code Formatter

Scowluga:
# Background

This PR introduces a code formatter, translating `ASTNode` to Alda code.
```
func FormatASTToCode(root ASTNode, out io.Writer, opts ...formatterOption) error {
```

# Notes
### Whitespaces
The formatter is quite rudimentary, but has basic handling of whitespacing (indents, spaces, line wrapping) while maintaining the integrity of the generated code.
- For indents, we essentially maintain an indentation level. Within parts, voices, and event seqs, we indent.
- For wrapping, we format "texts" or tokens that are single strings. Between texts, we can wrap the line.

### Variable Definitions
Variable definitions are tricky as I've noted in code comments. As a summary, the problem is all definition nodes must be on the same line as the variable name. The easiest way I found to handle this is to introduce the `varDefState` enum, which introduces some significant complexity into the formatter.

### Testing
We piggyback and use the `parseTestCase` to also automatically test formatting. Like with ast-gen, we ensure the "round-trip" formatted then parsed AST is identical to the actual correct AST.

### Comments
Currently comments are dropped when scanning, so they will be dropped for a round-trip format (code -> AST -> code).

I've spent a couple hours trying my hand at the changes necessary to retain comments to the AST level. I think this change is not worth it, as it's quite complicated and touches sensitive code. Plus MusicXML import doesn't produce comments, so the change is not necessary for this main use case. Plus the formatter itself is not very good, so I don't think it'll really standalone.

I've left supporting comments as a TODO.

### Commands
The formatter is not quite good enough in my opinion to standalone as it's own command to actually use. Combined with the fact that we don't even support comments, we will introduce an experimental hidden `alda format` command.

### DIFF
--- a/client/cmd/format.go
+++ b/client/cmd/format.go
@@ -1,0 +1,131 @@
+package cmd
+
+import (
+	"alda.io/client/color"
+	"alda.io/client/help"
+	log "alda.io/client/logging"
+	"alda.io/client/parser"
+	"fmt"
+	"github.com/spf13/cobra"
+	"io"
+	"os"
+)
+
+var formatInputFile string
+var formatOverwrite bool
+var formatConfiguredWrapLen int
+var formatConfiguredIndentText string
+
+func init() {
+	formatCmd.Flags().StringVarP(
+		&formatInputFile, "file", "f", "", "Input Alda file to format",
+	)
+
+	formatCmd.Flags().BoolVarP(
+		&formatOverwrite, "output", "o", false, "Overwrite input file with formatted output",
+	)
+
+	formatCmd.Flags().IntVarP(
+		&formatConfiguredWrapLen, "wrap", "w", 0, "Configured line character length to wrap formatted output (default 80)",
+	)
+
+	formatCmd.Flags().StringVarP(
+		&formatConfiguredIndentText, "indent", "i", "", "Configured indent text (default two spaces)",
+	)
+}
+
+var formatCmd = &cobra.Command{
+	Use:    "format",
+	Hidden: true,
+	Short:  "Format Alda source code",
+	Long: `Format Alda source code
+
+---
+
+Source code must be provided by specifying the path to a file (-f, --file):
+  alda format -f path/to/my-score.alda
+
+In this case, the formatted output will be printed to standard output.
+
+When -o / --output is specified, the input file is instead overwritten.
+  alda format -f path/to/my-score.alda -o
+
+Formatted output can be configured with the -w / --wrap and -i / --indent flags.
+  alda format -f path/to/my-score.alda -w 120 -i "    "
+
+---
+
+Currently, formatting cannot handle comments (i.e. all comments are dropped)
+
+---`,
+	RunE: func(_ *cobra.Command, args []string) error {
+		// TODO (experimental): remove warning log
+		log.Warn().Msg(fmt.Sprintf(
+			`The %s command is currently experimental. All comments are dropped during formatting.`,
+			color.Aurora.BrightYellow("format"),
+		))
+
+		root, err := parser.ParseFile(formatInputFile)
+		if err != nil {
+			return err
+		}
+
+		var out io.Writer
+		if formatOverwrite {
+			f, err := os.OpenFile(
+				formatInputFile,
+				os.O_WRONLY|os.O_TRUNC,
+				0664, // default rw-rw-r perms
+			)
+			if err != nil {
+				return help.UserFacingErrorf(
+					`Issue opening file %s.`,
+					color.Aurora.BrightYellow(outputAldaFilename),
+				)
+			}
+			defer f.Close()
+			out = f
+		} else {
+			out = os.Stdout
+		}
+
+		if formatConfiguredWrapLen < 0 {
+			return help.UserFacingErrorf(
+				`Configured line wrap length %d must be positive.`,
+				formatConfiguredWrapLen,
+			)
+		}
+
+		if formatConfiguredWrapLen > 0 && len(formatConfiguredIndentText) > 0 {
+			err = parser.FormatASTToCode(
+				root,
+				out,
+				parser.ConfigureSoftWrapLen(formatConfiguredWrapLen),
+				parser.ConfigureIndentText(formatConfiguredIndentText),
+			)
+		} else if formatConfiguredWrapLen > 0 {
+			err = parser.FormatASTToCode(
+				root,
+				out,
+				parser.ConfigureSoftWrapLen(formatConfiguredWrapLen),
+			)
+		} else if len(formatConfiguredIndentText) > 0 {
+			err = parser.FormatASTToCode(
+				root,
+				out,
+				parser.ConfigureIndentText(formatConfiguredIndentText),
+			)
+		} else {
+			err = parser.FormatASTToCode(root, out)
+		}
+
+		if err != nil {
+			return help.UserFacingErrorf(
+				`Issue formatting Alda: %s.`,
+				err.Error(),
+			)
+		}
+
+		return nil
+	},
+}
--- a/client/cmd/format.go
+++ b/client/cmd/format.go
@@ -22,7 +22,7 @@ func init() {
 	)
 
 	formatCmd.Flags().BoolVarP(
-		&formatOverwrite, "output", "o", false, "Overwrite input file with formatted output",
+		&formatOverwrite, "overwrite", "o", false, "Overwrite input file with formatted output",
 	)
 
 	formatCmd.Flags().IntVarP(
@@ -47,7 +47,7 @@ Source code must be provided by specifying the path to a file (-f, --file):
 
 In this case, the formatted output will be printed to standard output.
 
-When -o / --output is specified, the input file is instead overwritten.
+When -o / --overwrite is specified, the input file is instead overwritten.
   alda format -f path/to/my-score.alda -o
 
 Formatted output can be configured with the -w / --wrap and -i / --indent flags.
--- a/client/parser/format.go
+++ b/client/parser/format.go
@@ -152,34 +152,6 @@ func (f *formatter) formatWithDuration(
 
 			text.Reset()
 
-		case NoteLengthMsNode:
-			if shouldTie {
-				text.WriteString("~")
-			}
-
-			// NoteLengthMs is parsed to us as ms, but can be coded using s/ms
-			// We will try our best to use seconds first
-			totalMs := child.Literal.(float64)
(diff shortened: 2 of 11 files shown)


client/parser/format.go
```diff
@@ -0,0 +1,737 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"strconv"
+	"strings"
+)
+
+type varDefState int
```
Scowluga, line 11: @daveyarwood  This is the main piece of complexity that I think makes the formatter a bit spaghetti. I couldn't think of a better way to do it though. The fact is, it properly encapsulates the logic of what we want. Only during the last node of a var def should we allow indentation to wrap and continue onto the next line. It's a bit tricky, but I think it's safe enough for our purposes.

client/cmd/format.go
```diff
@@ -0,0 +1,87 @@
...
+	Hidden: true,
+	Short: "Format Alda source code",
+	Long: `Format Alda source code
+
+---
+
+Source code must be provided by specifying the path to a file (-f, --file):
+  alda format -f path/to/my-score.alda
+
+In this case, the score will be formatted in-place (i.e. overwritten).
```
daveyarwood, line 36: To make this formatter more flexible, I think it should print the result to stdout by default, and we should include a flag to enable overwriting the file.

`sed` and [`gofmt`](https://pkg.go.dev/cmd/gofmt) both operate this way -- they write the result to stdout by default, and there is a flag (`sed -i`, `gofmt -w`) to opt into overwriting the file.

client/parser/test_helper.go
```diff
@@ -73,6 +74,23 @@ func executeParseTestCases(t *testing.T, testCases ...parseTestCase) {
...
+		}
+		formattedAST, err := Parse(
+			// The newly formatted file will have different source context's
+			testCase.label, buffer.String(), SuppressSourceContext)
+		if diff := deep.Equal(actualAST, formattedAST); diff != nil {
+			t.Error(testCase.label)
+			for _, diffItem := range diff {
+				t.Errorf("%v", diffItem)
+			}
+		}
```
daveyarwood, line 92: Nice idea, testing that the formatted code parses to an equivalent AST as the original code :+1:

client/parser/format.go
```diff
@@ -0,0 +1,748 @@
...
+	out         io.Writer
+}
+
+type formatterOption func(*formatter)
+
+func newFormatter(out io.Writer, opts ...formatterOption) *formatter {
+	formatter := &formatter{
+		// TODO: make these configurable from cli
+		softWrapLen: 80,
+		indentText:  "    ",
```
daveyarwood, line 34: For now, let's make this 2 spaces by default. Personal preference :)   To your point above, we could eventually make this configurable.

client/parser/format.go
```diff
@@ -0,0 +1,748 @@
...
+
+			text.Reset()
+
+		case NoteLengthMsNode:
+			if shouldTie {
+				text.WriteString("~")
+			}
+
+			// NoteLengthMs is parsed to us as ms, but can be coded using s/ms
+			// We will try our best to use seconds first
```
daveyarwood, line 150: Wouldn't this result in a potentially different AST than the original code? It seems like to keep things simple, it would be better if we just formatted this as milliseconds.

Is the issue here that the parser normalizes seconds into a NoteLengthMsNode, so at this point, we don't know if the original code was seconds or milliseconds? If so, I don't think it would be a problem to add a NoteLengthSecondsNode, so that we can preserve the original intention.

client/parser/test_helper.go
Scowluga: Thanks, I think this invariant should be maintained in all cases....

There is one interesting thing I realized which is when transforming `model.AttributeUpdate{model.PartUpdate}` back into Lisp for formatting, we have to arbitrarily pick one of possibly multiple Lisp defn's (e.x. "transposition" vs "transpose"). This doesn't affect testing the formatter though as we don't go that deep in this parser test case.

client/parser/format.go
```diff
@@ -0,0 +1,748 @@
...
+			f.write("]")
+
+		case LispListNode:
+			var lispString func(ASTNode) (string, error)
+			lispString = func(lisp ASTNode) (string, error) {
+				switch lisp.Type {
+
+				default:
+					return "", fmt.Errorf(
+						"unexpected LispLispNode %#v during formatting", lisp,
```
daveyarwood, line 319: Is this a typo?

client/parser/format.go
```diff
@@ -0,0 +1,748 @@
...
+
+		case VariableDefinitionNode:
+			// Variable definitions are incredibly tricky to formatInnerEvents because
+			// formatted text must be on the same line as the variable name.
+			// We handle this by maintaining varDefState:
+			// - While "None", behaviour is normal.
+			// - While "Defining", we never flush/wrap to a new line.
+			// - While "LastNode", we allow event sequences (including repeats)
+			// 	 and voice groups to indent and continue on new lines.
+			// 	 This is complicated, but any alternative I tried was worse.
```
daveyarwood, line 547: I can't think of any alternatives at the moment, but I think this is perfectly fine for now.

daveyarwood (commented):
Overall, this looks great. Nice work!

client/parser/format.go
Scowluga: Ah I thought I was doing something clever, but this is just an oversight from me.

So yeah the scanner just produces the literal as a ms float64. So "1s" turns into "1000". 

I will make the change to add a `NoteLengthSecondsNode` and equivalents then. So we can preserve the original intention.

client/parser/test_helper.go
daveyarwood: Hmm, this seems like a hard problem, because the idea with the Lisp code is that it evaluates code into a ScoreUpdate. But we can't accurately turn a ScoreUpdate back into the code that produced it, because there can be many different code expressions that evaluate to the same ScoreUpdate. This thing about attribute aliases (e.g. "vol" vs. "volume") is a very simple example, but theoretically in the future, we could decide to add more built-in functions to the Lisp language, which could make it all the more difficult (impossible, actually) to generate the correct code from a ScoreUpdate.

Imagine, for example, that we added math operators, so that you could express `(vol 100)` in many ways:

```lisp
(vol (+ 50 50))
(vol (* 10 10))
(vol (- 200 100))
```

I think it would be problematic if the formatter simplified Lisp code expressions like the above as `(vol 100)`, or even `(volume 100)`, because the job of a formatter isn't to evaluate or simplify code, it's to format it.

I don't think we necessarily need to do anything about this right now (maybe a FIXME/TODO comment would suffice), but one idea of a potential solution is to add an optional Metadata property on all ScoreUpdates, and store the Lisp expression there after we evaluate the expression, so that we can retain it for the purposes of converting back to code. Actually, this Metadata property might help us in other contexts as well, whenever we have any kind of situation where multiple possible AST nodes could produced the same ScoreUpdate.

daveyarwood: Importantly, the Metadata would be ignored when doing deep diffs.

I got this metadata idea from Clojure. Clojure lets you attach metadata to values, and the metadata can be inspected as needed, but otherwise, the values work exactly the same way as they always do:

```clojure
user=> (def a ^{:my-property "my value"} {:x 42})
#'user/a
user=> (def b ^{:my-property "different value"} {:x 42})
#'user/b
user=> a
{:x 42}
user=> b
{:x 42}
user=> (meta a)
{:my-property "my value"}
user=> (meta b)
{:my-property "different value"}
;; a and b are considered equal, despite having different metadata
user=> (= a b)
true
```

client/parser/format.go
Scowluga: Changed to 2 spaces default. Also updated the cmd to make those two configurable.

client/cmd/format.go
Scowluga: Good suggestion. Done. Now `-f` specifies input file, default printing to `os.Stdout`. Specifying `-o` will overwrite the input file.

client/parser/format.go
Scowluga: Made changes here: https://github.com/alda-lang/alda/pull/474/commits/902938c74bb72bb12fe2b303ab7497acc496d65f.

I essentially added 
- `NoteLengthSeconds` token to the scanner
- `NoteLengthSecondsNode` node to the parser
- `model.NoteLengthSeconds` score update to model

Leaving to let you resolve.

client/parser/test_helper.go
Scowluga: I'm a bit confused, but I'm thinking we're ok right now and it seems we don't have to do anything.
- For `alda format`, we never actually go past the `ASTNode` layer to `ScoreUpdate`s. This means the Lisp is never evaluated or simplified. So we should be good here. The point of AST I guess is to only be the structure of the language.
- For `alda import`, we generate whatever ScoreUpdates we want. Right now, every `ScoreUpdate` can be generated to `ASTNode`. As long as we cover this single use case of code generation, we're good for now.
- Down the line if we run into issues, we can make changes as necessary.

What TODO exactly are you trying to point out? I can see how Lisp eval can be "lossy" of the original code, but that's not a problem with formatting because we only go to AST. Perhaps there can be issues down the line generating code from complicated lisp `ScoreUpdate`s, but surely we can just generate something reasonable.

Sorry if I'm misunderstanding, this Lisp part has always confused me.

client/cmd/format.go
```diff
@@ -0,0 +1,131 @@
...
+var formatConfiguredWrapLen int
+var formatConfiguredIndentText string
+
+func init() {
+	formatCmd.Flags().StringVarP(
+		&formatInputFile, "file", "f", "", "Input Alda file to format",
+	)
+
+	formatCmd.Flags().BoolVarP(
+		&formatOverwrite, "output", "o", false, "Overwrite input file with formatted output",
```
daveyarwood, line 25: Could we change the name of this flag to `--overwrite`? I think that would be clearer.

client/parser/test_helper.go
daveyarwood: Sorry David, I think I confused both of us! I'm thinking about it more now, and I think you're right that this is not an issue. It doesn't look like we would ever need to go from AST -> ScoreUpdates -> AST -> formatted code. Disregard my rambling above :smile:

daveyarwood:
@Scowluga I think this PR is just about ready to merge -- let's just rename the `--output` flag to `--overwrite`

daveyarwood (approved):

PR #475, opened 2023-03-26
Complete MusicXML Import

Scowluga:
# Background

This PR finally finishes the `alda import` command and marks it as no longer hidden / experimental.

# Command

### score-partwise
As mentioned in code comments, I considered looking into applying the XSLT stylesheet directly in code to ensure score-timewise MusicXML files are converted automatically to score-partwise. Turns out nobody really uses score-timewise, so it's not quite worth the additional dependencies.

### MusicXML version
We now log a warning for MusicXML versions that are not 3.1. The current new version is 4, which most people should hopefully be using. The warning message may get annoying, but I don't think I'll have sufficient time to ensure import is working with version 4. Furthermore, in the future, there may be even newer (and even older) versions that just mess everything up. Better to warn.

# Importer Changes
**We are only trying to achieve a baseline importer that gets you 90% of the way to a complete Alda score**.

There will be various issues with importing, especially when handling edge cases and complex MusicXML files.

I made various additional commits improving the MusicXML importer to get to what I believe is a fair baseline correctness. These are in separate commits with descriptive commit messages.


client/cmd/import.go
```diff
@@ -37,8 +37,8 @@ func init() {
 
 var importCmd = &cobra.Command{
 	Use:    "import",
-	Short:  "Evaluate external format and import as Alda source code",
-	Long: `Evaluate external format and import as Alda source code
+	Short:  "Import Alda source code from MusicXML",
+	Long: `Import Alda source code from MusicXML
```
daveyarwood, line 41: Suggestion: "Import Alda source code from other formats"

While it's true that MusicXML is currently the only supported format, I think the help text should describe the general intent of this command, which is to potentially support various formats in the future.

client/interop/musicxml/importer/importer.go
```diff
@@ -353,25 +353,37 @@ func ImportMusicXML(r io.Reader) ([]model.ScoreUpdate, error) {
...
 	}
 
+	version := scorePartwise.SelectAttrValue("version", "")
+	if version != "3.1" {
+		log.Warn().Msg(fmt.Sprintf(
+			`MusicXML import supports version %s. Behaviour may be undefined with provided version %s.`,
+			color.Aurora.BrightYellow("3.1"),
+			color.Aurora.BrightYellow(version),
+		))
+	}
```
daveyarwood, line 385: I agree that it's good to log a warning, since we only support 3.1 at the moment :+1:

daveyarwood (commented):
Looks good!

client/interop/musicxml/importer/util.go
```diff
@@ -330,3 +336,68 @@ func isOrContainsBarline(update model.ScoreUpdate) bool {
 
 	return false
 }
+
+// idiomaticDuration applies simple heuristics to make durations more idiomatic
+func idiomaticDuration(aldaDuration float64, dots int32) model.Duration {
```
daveyarwood, line 341: Nice work on this part, in particular. This looks like it was hard to write, but it's totally worthwhile!

daveyarwood (approved):
This is awesome! I'm very very happy with this :raised_hands:

PR #478, opened 2023-04-06
Miscellaneous MusicXML Bugs

Scowluga:
Had some more time to do some smoke testing, so fixed two more things. See commits.

Can just merge.


client/interop/musicxml/importer/util.go
```diff
@@ -401,3 +403,12 @@ func idiomaticDuration(aldaDuration float64, dots int32) model.Duration {
 		}},
 	}
 }
+
+// round rounds values to deal with floating point arithmetic issues
+func round(val float64) float64 {
```
Scowluga, line 408: I think this deserves a better name but I can't quite put my finger on one...

daveyarwood (approved):
Looks good - thanks!

PR #490, opened 2023-10-13
refactor: remove io/ioutil

ginglis13:
io/ioutil has been deprecated since Go 1.16
https://pkg.go.dev/io/ioutil


daveyarwood:
Looks good. All tests and `alda doctor` tests are passing. Thank you for the PR!

Issue #405, opened 2021-10-22
CLI output is wonky


elyisgreat:
# 🐞 Bug report 🐞

## Description
I'll let the below image speak for itself:

![Screenshot (22)](https://user-images.githubusercontent.com/8622264/138392256-2e5aa553-83da-4576-8be3-57f943d3b33f.png)

Running the REPL causes PowerShell to recolour the screen in a weird way. Running this on cmd doesn't cause any recolouring, but you still get these `←[0m` and `←[36m` strings appearing.

grasinski:
Hi, I'd be willing to look into this!

KenP97:
Hi, can i look into this too?

daveyarwood:
Great, thanks!

KenP97:
I'm guessing the issue to be with the readline part, I changed the color library to "github.com/grookit/color" and managed to fix those ANSI escape characters printing out... but when the readline part starts the problems persists, and the background changes.

And I tried running readline separately and still the issue persisted even on a new project. so I'm guessing the issue is with the readline (we might need to configure it because they did mention it works on PowerShell)... ill look more into it till then
![image](https://user-images.githubusercontent.com/19164140/168334227-8f97fc45-1647-4709-a9d4-a96addab151a.png)

KenP97:
Am I allowed to use a different library instead of readline?

KenP97:
![image](https://user-images.githubusercontent.com/19164140/168538589-71164a46-d9ab-4108-9213-039c8bccffaa.png)

daveyarwood:
I answered you on Slack already, but for visibility here:

* Changing libraries is definitely an option.
* Changing the `log.SetOutput(...)` part could be an issue.  I think I added that part to make sure that log messages don't interfere with inputting lines in the REPL. So whatever library we switch to, we should just make sure that log messages don't interfere with the REPL experience.
* If you could make a PR, I'll be happy to test on Linux!

PR #501, opened 2024-07-27
PS color output fix

Vanello1908:
Fixed one of bugs mentioned in [this](https://github.com/alda-lang/alda/issues/405#issue-1033159114) issue

### DIFF
--- a/client/color/color.go
+++ b/client/color/color.go
@@ -2,9 +2,11 @@ package color
 
 import (
 	"os"
+	"runtime"
 
 	auroraLib "github.com/logrusorgru/aurora"
 	"github.com/mattn/go-isatty"
+	"golang.org/x/sys/windows/registry"
 )
 
 // We're using a couple of libraries that produce ANSI escape sequences to print
@@ -40,6 +42,20 @@ func init() {
 	// HACK: Ideally, aurora would support NO_COLOR, but at least they give us a
 	// config option so that we can disable color manually.
 	//
-	// See the longer comment above EnableColor.
+	if runtime.GOOS == "windows" {
+		var key, err = registry.OpenKey(registry.CURRENT_USER, "Console", registry.QUERY_VALUE)
+		if err != nil {
+			EnableColor = false
+		} else {
+			var val, _, err = key.GetIntegerValue("VirtualTerminalLevel")
+			if err != nil {
+				EnableColor = false
+			} else if val == 0 {
+				EnableColor = false
+			}
+		}
+	}
+	print("Enable color: ")
+	print(EnableColor)
 	Aurora = auroraLib.NewAurora(EnableColor)
 }
--- a/client/color/color.go
+++ b/client/color/color.go
@@ -41,7 +41,8 @@ var Aurora auroraLib.Aurora
 func init() {
 	// HACK: Ideally, aurora would support NO_COLOR, but at least they give us a
 	// config option so that we can disable color manually.
-	//
+
+	//Check registry for enabled color printing(only for windows)
 	if runtime.GOOS == "windows" {
 		var key, err = registry.OpenKey(registry.CURRENT_USER, "Console", registry.QUERY_VALUE)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


daveyarwood:
Hi, thanks for the PR! I tried this branch locally and unfortunately, it doesn't compile on Linux. I get this error:

```
$ go build .
package alda.io/client
        imports alda.io/client/cmd
        imports alda.io/client/color
        imports golang.org/x/sys/windows/registry: build constraints exclude all Go files in /home/dave/.go/pkg/mod/golang.org/x/sys@v0.22.0/windows/registry
```

I did some googling and found [this StackOverflow question](https://stackoverflow.com/questions/72849222/golang-1-18-exclude-package-for-specific-platform) which looks like the same issue.

If we wanted to get this approach to work, I think we would need to use [build constraints](https://pkg.go.dev/go/build#hdr-Build_Constraints) to exclude the Windows registry code when the code is being compiled and run in a non-Windows environment.

I did a little test of this by copy-pasting the code you added into a new file called `color_windows.go`, and it did build successfully after that. However, it won't exactly work that way, because there's a potential race condition where we could end up running `Aurora = auroraLib.NewAurora(EnableColor)` in `color.go` before the code in `color_windows.go` sets `EnableColor` appropriately.

If you'd like to keep working on this approach and adjust it along those lines so that it will compile on Linux, you're welcome to do so. You can check the CircleCI logs to see if the build succeeded, i.e. click the :x: or :heavy_check_mark: next to the commit message:

2024-07-27-173632_1156x324_scrot

And then click the "Details" link to go to the build logs in CircleCI:

2024-07-27-173704_1454x852_scrot

daveyarwood:
My other thought on this is that it would be better if we didn't have to include a bunch of ad hoc code in Alda to work around OS-specific issues. We should really let the color library handle this for us.

The version of the color library we're using (Aurora) is from 2020, which is several years old at this point. I think it's worth trying to update to the latest version and see if it works better out of the box in Powershell.

Vanello1908:
Ok, this is done with build constraint. Next step - black symbols, it is a problem with readline lib. There are 3 options:
- create pull request to that lib with fix
- create fork with fix
- use another lib for input
But I want not to use the third option

client/color/color_widows.go
```diff
@@ -0,0 +1,63 @@
+//go:build windows
```
daveyarwood, line 1: I think you can remove this line. The Go compiler recognizes when the file name ends with `_windows.go` and only includes it in the build if the OS is Windows.

(You also have a typo in the filename -- should be `color_windows.go`)

client/color/color.go
```diff
@@ -39,7 +41,5 @@ var Aurora auroraLib.Aurora
 func init() {
 	// HACK: Ideally, aurora would support NO_COLOR, but at least they give us a
 	// config option so that we can disable color manually.
-	//
-	// See the longer comment above EnableColor.
```
daveyarwood, line 43: Let's put this comment back, as it's still relevant.

client/color/color_widows.go
```diff
@@ -0,0 +1,63 @@
...
+	len(os.Getenv("NO_COLOR")) == 0
+
+var Aurora auroraLib.Aurora
+
+func init() {
+	// HACK: Ideally, aurora would support NO_COLOR, but at least they give us a
+	// config option so that we can disable color manually.
+
+	//Check registry for enabled color printing(only for windows)
+	if runtime.GOOS == "windows" {
```
daveyarwood, line 48: We probably don't need this check, as this file will only be included in the Windows build.

daveyarwood:
Looking good so far! If you don't mind implementing my suggestions above, I think this will be ready to merge.

I am curious if you've tried updating the Aurora dependency, to see if maybe the library has improved in the last 4 years to the point that none of this OS-specific code is necessary. But I'm also fine with the approach in this PR, at least as an interim improvement.

Vanello1908:
image
aurora v2.0.3+incompatible also has this problems, so some cleanup and last commit :)

Vanello1908:
Aurora v2.0.3+incompatible also has problems with PowerShell, so we remain this code

daveyarwood (approved):
Looks good! I'm going to go ahead and merge this and cut a patch release of Alda. I appreciate your work on this, and will be happy to review any future PRs to address the remaining Windows color issues.

Issue #416, opened 2022-04-16
Alda REPL server sometimes "claims" more than 1 player on startup


daveyarwood:
# 🐞 Bug report 🐞

## Description

When starting an Alda REPL session, the REPL server finds an available player process (i.e. one in the "ready" state) and "claims" it by sending it a ping message. This puts the player process into the "active" state, which makes Alda more eager to spawn new player processes in order to have them readily available.

This works reliably, but it seems to be a little _too_ reliable in that, for some reason, _two_ player processes seem to go from "ready" to "active" when the REPL server starts.

## Steps to Reproduce

1. In a separate terminal, monitor the state of the player processes by running `watch alda ps`.
2. Start from a clean state by running `alda shutdown` to shut down all current player processes.
3. Run `alda -v2 repl` to start a REPL session in verbose mode.
4. Exit the session and start a new one.
5. Repeat step 4.

## Expected Behavior

Each time a REPL session is started, only 1 of the players in the "ready" state goes into the "active" state.

Alda spawns more player processes as needed, but it shouldn't need to do that too much, because only one process becomes unavailable for each REPL session.

## Actual Behavior

The first time I start a REPL server session, I see 5 players in the "ready" state and 1 player in the "active" state. The REPL server output shows that it is using the 1 player in the "active" state.

When I exit that session and start another one, 2 of the players that were in the "ready" state go into the "active" state. One of them is the one that this new REPL session is using. We are now down to 3 players in the "ready" state.

If I keep going, exiting and starting new REPL sessions, more and more player processes are spawned to ensure that there are always player processes available in the "ready" state.

## Environment

**Operating system and version:** Ubuntu 20.04

**Alda version:**

```
$ alda version
alda 2.2.1
```

```
$ alda-player info
alda-player 2.2.1
log path: /home/dave/.cache/alda/logs
```

**Health check:**

```
$ alda doctor
(all passing)
```

PR #507, opened 2025-03-18
fix issues with excess active processes

Gooding706:
Hello Dave,
I fixed the issue mentioned in issue #416. 

The issue was related to processes remaining in the active state even after they were no longer being utilized (i.e when the repl server was killed). to fix this I added the /yield command which simply sets the process to the ready state acting as the inverse of the /ping command. Just this alone didn't fix the whole problem as there was a little bug with the printing of the "Starting player processes..." message where a player process would be pinged and set to the active state purely to check if any processes exist. This issue was resolved by slightly altering the implementation of that functionality. 

Let me know if there is anything I should change.

### DIFF
--- a/client/system/process_management.go
+++ b/client/system/process_management.go
@@ -589,7 +589,7 @@ func FillPlayerPool() error {
 // processes come up, we print a message to make it clear what we're waiting
 // for.
 func StartingPlayerProcesses() {
-	if _, err := FindAvailablePlayer(); err == ErrNoPlayersAvailable {
+	if states, _ := ReadPlayerStates(); len(states) == 0 {
 		fmt.Fprintln(os.Stderr, "Starting player processes...")
 	}
 }
--- a/client/repl/client.go
+++ b/client/repl/client.go
@@ -553,6 +553,11 @@ func (client *Client) handleCommand(name string, args string) error {
 // Disconnect closes the client's connection with the server.
 func (client *Client) Disconnect() {
 	if client.serverConn != nil {
+		_, err := client.sendRequest(map[string]interface{}{"op": "yield"})
+		if err != nil {
+			fmt.Fprint(os.Stderr, "request to yield player proccess failed")
+		}
+
 		client.serverConn.Close()
 	}
 }
--- a/client/repl/server.go
+++ b/client/repl/server.go
@@ -531,6 +531,19 @@ var ops = map[string]func(*Server, nREPLRequest){
 
 		server.respondDone(req, nil)
 	},
+
+	"yield": func(server *Server, req nREPLRequest) {
+		if err := server.withTransmitter(
+			func(transmitter transmitter.OSCTransmitter) error {
+				return transmitter.TransmitYieldMessage()
+			},
+		); err != nil {
+			server.respondError(req, err.Error(), nil)
+			return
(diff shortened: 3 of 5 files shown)


daveyarwood:
Hi @Gooding706, thank you for looking into this!

One issue here is that it isn't necessarily safe to change a player process's status from "active" to "ready". If a player process is active, it means the state of the MIDI sequencer and synthesizer are potentially changing and are no longer the initial "blank slate" that is expected when a player process is in the "ready" state.

Gooding706:
@daveyarwood, thanks for checking out my request, thats a good point I hadn't thought of. I would assume that adding `systemActions.add(SystemAction.CLEAR)` to the yield command may have some effect in putting the player into a clean state although I'm not 100% sure. Another solution would just be to shutdown the active process at the end of a repl session which has the disadvantage that we would lose a process, however, in the current implementation we already lose processes on shutdown as they end up inaccessible in the active state.

daveyarwood:
Shutting down or reverting the active process at the end of a REPL session is not really a problem. Player processes eventually shut themselves down after a period of inactivity, so the active process will go away on its own.

The issue I was talking about in #416 is that there seems to be a bug that causes the REPL server to put _not 1, but 2_ player processes into the "active" state, when the REPL server starts up.

Gooding706:
I must have misunderstood, the reason that is happening is due to the way `StartingPlayerProcesses()` behaves when there are available processes due to its use of `FindAvailablePlayer()`. I'll can submit a commit with just that being addressed. Sorry about the confusion.

client/system/process_management.go
```diff
@@ -589,7 +589,7 @@ func FillPlayerPool() error {
 // processes come up, we print a message to make it clear what we're waiting
 // for.
 func StartingPlayerProcesses() {
-	if _, err := FindAvailablePlayer(); err == ErrNoPlayersAvailable {
+	if states, _ := ReadPlayerStates(); len(states) == 0 {
```
daveyarwood, line 592: I think you're on the right track here, but it isn't exactly ideal this way, either.

The problem is that `FindAvailablePlayer()` not only checks that there is at least 1 player process (i.e. `ReadPlayerStates()` returns at least one state), but it also checks that:

* There is at least 1 player process in the "ready" state, and
* That player responds to a ping message

So `FindAvailablePlayer()` is a good thing to do, but the thing is, we're already calling it anyway in all of the places where we're calling `StartingPlayerProcesses()` first. The only purpose of `StartingPlayerProcesses` is to print the "Starting player processes..." message when we fill the player pool.

So, thinking about this more now, I'm thinking that we should just get rid of the `StartingPlayerProcesses` function, and instead, `FindAvailablePlayer()` should print the "Starting player processes..." message when it fills the player pool. I'm 90% sure that this will behave exactly the same as it does currently, except that we won't have excess active processes.

daveyarwood: The only thing that will be a little different is that the REPL _server_ will now print "Starting player processes..." instead of the REPL _client_, so the timing of the message being printed will be a little different than what it is currently when you run `alda repl`. I'm tentatively OK with that, we just need to see what it looks like and make sure it doesn't look totally weird, e.g. interleaved with other output.

PR #508, opened 2025-05-19
Nested repeat fix

### LINKED ISSUE #401: Repetition variations don't behave well in a sequence containing repeats
# 🐞 Bug report 🐞

## Description

We are storing a single `currentRepetition` value (see repeat.go and repetitions.go) and using it to track the current repetition number that we are going through a repeated sequence of events.

However, a repeated sequence can itself contain repeated events, and the nesting can be arbitrarily deep.

I think it would probably make sense to refactor so that instead of a single "current repetition" value, we are working with a stack of "current repetition" values, where we use the topmost one and discard it when we exit the loop.

## Steps to Reproduce

Repro case, adapted from the one @wesen provided on Slack:

```alda
period = r16
motif = a b c

(tempo! 100)

piano:
  V1: o4 [{motif}8 period period <'2 ]*4
```

This behaves as expected. The motif is played twice in octave 4, then twice in octave 3, because on the second repetition, the "octave down" event is applied.

```alda
period = r16
motif = a b c

(tempo! 100)

piano:
  V1: o4 [{motif}8 period*2 <'2 ]*4
```

This _should_ behave the same, however the "octave down" event ends up getting applied on _every_ repetition, so we play the motif in octave 4, then 3, then 2, then 1.

## Environment

**Alda version:**

```
$ alda version
alda 2.0.5
```

```
$ alda-player info
alda-player 2.0.5
log path: /home/dave/.cache/alda/logs
```

De-Alchmst:
This should fix #401.

I have taken different approach than suggested, as it was more time efficient and less error-prone than trying to handle it from outside. Allocating a small array per repeat should not affect performance in any noticeable way.

### DIFF
--- a/client/model/repeat.go
+++ b/client/model/repeat.go
@@ -31,13 +31,20 @@ func (repeat Repeat) JSON() *json.Container {
 // UpdateScore implements ScoreUpdate.UpdateScore by repeatedly updating the
 // score with an event a specified number of times.
 func (repeat Repeat) UpdateScore(score *Score) error {
+	previousRepetitions := make([]int32, len(score.CurrentParts))
+
 	for repetition := int32(1); repetition <= repeat.Times; repetition++ {
-		for _, part := range score.CurrentParts {
+		for i, part := range score.CurrentParts {
+			previousRepetitions[i] = part.currentRepetition
 			part.currentRepetition = repetition
 		}
 
 		if err := score.Update(repeat.Event); err != nil {
 			return err
+		}
+
+		for i, part := range score.CurrentParts {
+			part.currentRepetition = previousRepetitions[i]
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


daveyarwood:
@De-Alchmst Thanks so much for the PR! I'm in the middle of moving right now and a little busy, but I'm hoping to review this in more detail. I did take a quick look, and it looks like a straightforward approach, I just want to make sure it definitively fixes the issue.

In the meantime, would you be up for writing a test case for the issue described in #401 and ensuring that it passes?

De-Alchmst:
I'll try looking into it.

daveyarwood:
Looks great! Thanks for the tests, as well. All tests are passing, and I also tested the example in #401 and confirmed that the issue is fixed :tada:

PR #512, opened 2025-10-06
Fixed accumulating delay. Added offset Log Debug messages.

### LINKED ISSUE #415: Alda REPL: "Play right away" offset calculation doesn't work with voices
# 🐞 Bug report 🐞

## Description

In the Alda REPL use case, the user is appending additional events to the score, and as they're doing that, we want to play the new notes immediately, not wait until they would come up in the score. There shouldn't be any delay between entering a line of code in the REPL and hearing it.

To support this use case, [we play some tricks](https://github.com/alda-lang/alda/blob/798e6df8533db0258485856f92757f445ebb3455/client/transmitter/osc.go#L342-L352) in the note scheduling code, where the caller can pass in a map of "sync offsets" for each part, and we will subtract the offset for each part from all of its notes. As a result, we end up with notes that would be played right away, not however far into the score they would be if this were an Alda score file and not a REPL session.

This has been working great, but I just noticed that for some reason, it doesn't seem to work if you use voices.

## Steps to Reproduce

1. Start a REPL session (`alda repl`)
2. Type in `piano: V1: o4 c d e  V2: o5 c d e` and press Enter.
3. Press the Up arrow to recall the previous line and press Enter to run it again.
4. Repeat step 3 several times.

## Expected Behavior

Every time you enter this line of code in the REPL, you hear the result immediately.

## Actual Behavior

The first time, you hear the result immediately.

The second time, the result is delayed by 3 beats.

The third time, the result is delayed by 6 beats, etc.

## Environment

**Operating system and version:** Ubuntu 20.04

**Alda version:**

```
$ alda version
alda 2.2.1
```

```
$ alda-player info
alda-player 2.2.1
log path: /home/dave/.cache/alda/logs
```

**Health check:**

```
$ alda doctor
(all passing)
```

**Logs:** not relevant here

JakubSlacht:
Fixes #415 

Hello.

This pull request should solve the accumulation of delay in a REPL session when using Voices. Voices still cause a slight delay compared to non-voice input, but the delay disappears once the current Voice input is ended/closed, which makes it a lot more usable. I've also added some debug logs useful for offset information in REPL sessions. I've tested this solution on the example provided in the issue and some other simple inputs.

Please, feel free to review and inform me if there are any errors with my fix. Thank you very much!

### DIFF
--- a/client/model/score.go
+++ b/client/model/score.go
@@ -162,7 +162,7 @@ func (score *Score) PartOffsets() map[*Part]float64 {
 	offsets := map[*Part]float64{}
 
 	for _, part := range score.Parts {
-		offsets[part.origin] = part.origin.CurrentOffset
+		offsets[part.origin] = part.CurrentOffset
 	}
 
 	return offsets
--- a/client/repl/server.go
+++ b/client/repl/server.go
@@ -569,6 +569,11 @@ func (server *Server) updateScoreWithInput(
 	// option when transmitting the score.)
 	partOffsets := server.score.PartOffsets()
 
+	// Log for checking what offset value we use for the current playback.
+	for name, offset := range partOffsets {
+		log.Debug().Str("part", name.Name).Float64("offset", offset).Msg("Offset before the update, taken into account.")
+	}
+
 	// Take note of the current `eventIndex` value, so that we know where to start
 	// playing from when we want to play the new events.
 	eventIndex := server.eventIndex
@@ -585,6 +590,11 @@ func (server *Server) updateScoreWithInput(
 
 	if err := server.score.Update(scoreUpdates...); err != nil {
 		return nil, err
+	}
+
+	// Log for checking what offset value we will use next playback. (Useful for tracking by how much the offset changed.)
+	for name, offset := range server.score.PartOffsets() {
+		log.Debug().Str("part", name.Name).Float64("offset", offset).Msg("Offset after the update, not yet taken into account.")
 	}
 
 	// Add the provided `input` to our total string of input representing the


daveyarwood:
@JakubSlacht Thanks so much for the PR, and sorry that I'm just now getting to this!

I tested this branch out and it's just like you said -- the delay is still there, but at least now it doesn't accumulate.

With the example from the issue (`piano: V1: o4 c d e  V2: o5 c d e`), it seems like there is always a delay of 3 beats (the length of the voice group). Ideally, I'd like to get to the root of the issue and fix it definitively so that there is no delay, regardless of whether or not you're using voices.

daveyarwood:
Hey @JakubSlacht - I played around with this myself for a bit and implemented a fix in a different way. In the process, I ended up making a few more changes to really nail it down, like making the part IDs stable for synchronization and tweaking some voice logic in the model – but your contributions were a critical first step, and I really appreciate it! Your initial digging and proposed solution were super helpful in getting me pointed in the right direction.

JakubSlacht:
Ah, I see. Thank you for your time and patience with me on this issue! Unfortunately, I do not think your solution fixes the bug completely? Because when I use the example scenario that I provided:

(each of these is a separate command)
`piano: V1: c d e`
`V1: c d e`
`V1: c d e`
`V2: o5 c d e`

The V2 section does not replay because the offset is actually further ahead than it should be. That's why I did that (admittedly very ugly) logic of determining the minimum offset between the voices. I am truly sorry to plague you with this issue still, but hopefully we can figure it out togehter?

daveyarwood:
@JakubSlacht No worries at all! Thanks for letting me know about this.

Do you think you could have a look at my code in the commit I linked above, and let me know what changes you think are needed in order to fix it for this scenario?

PR #519, opened 2026-03-08
Fix parseFloatFrom to use startIndex correctly

xoxwa:


daveyarwood (approved):
Nice catch. I don't think this is actually broken, since in practice, we only ever call `parseFloatFrom` with `s.start` as an argument, but using the `startIndex` argument was definitely the intention.

Thanks for the PR!

Issue #484, opened 2023-05-24
Lisp function `part` got deleted in Alda 2


UlyssesZh:
The following plays in Alda 1 but not in Alda 2:
```
(part "piano") c d e f
```
Errors in Alda 2:
```
<no file>:1:2 unresolvable symbol: part
```
It seems that `part` function is deleted without mentioning in the migration guide.

daveyarwood:
Good catch! I think we should add support for `part` in Alda 2 to be consistent.

Br0wnHammer:
Hey, is this issue still open to work on? If so, I would like to contribute.

daveyarwood:
@Br0wnHammer Yes, it's yours if you want it. 🙂

daveyarwood:
I just released Alda 2.4.2, which includes a basic version of the `part` function. (Thanks for the PR, @CodeForBeauty!). This supports the basic, one instrument use case, e.g. `(part "piano")`, just like in Alda 1.

One caveat: the current implementation only supports declaring a single instrument at a time. More advanced part declarations — multiple instruments in a group (e.g. `trumpet/piano`) and aliases (e.g. `'trumpiano'`) — are not yet supported via the part Lisp function -- that's a TODO for the future, if anyone wants to tackle it.

PR #520, opened 2026-03-16
Lisp function part defined.

CodeForBeauty:
Added a lisp function part as per this issue: #484 

The definition is added inside init function using defn, let me know if this is not how it should be done.

### DIFF
--- a/client/model/lisp.go
+++ b/client/model/lisp.go
@@ -1374,6 +1374,21 @@ func init() {
 			},
 		},
 	)
+
+	defn("part",
+		FunctionSignature{
+			ArgumentTypes: []LispForm{LispString{}},
+			Implementation: func(args ...LispForm) (LispForm, error) {
+				partName := args[0].(LispString).Value
+
+				partDecl := PartDeclaration{
+					Names: []string{partName},
+				}
+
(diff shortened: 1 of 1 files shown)


daveyarwood:
Hi @CodeForBeauty, thanks for the PR! The code looks good at first glance. I'm a little busy at the moment, but I'll test it soon and hopefully merge it.

daveyarwood (approved):
Thanks for this! The implementation looks correct for the simple case, and that's exactly what the original issue is asking for — `(part "piano") c d e f` works again, which is great.

I'm going to go ahead and merge this as-is. A couple of follow-up notes I'll take care of myself:

 - I'll add a test case in parts_test.go following the pattern used in `attributes_test.go` for other Lisp functions (e.g. "set octave using lisp").
 - I'll add a TODO comment in the implementation noting that `PartDeclaration` also supports multiple instrument names and an alias (possibly similar to [what I implemented in alda-clj](https://github.com/daveyarwood/alda-clj/blob/61ada6498c272c943d5c267e5d82066168ab0704/src/alda/core.clj#L268-L298)), which this Lisp function doesn't yet support. The API design for those cases deserves more thought, so I'm leaving it for a future contribution.

Thanks again for picking this up! 🎉

Issue #403, opened 2021-10-18
Alda on Raspberry Pi


daveyarwood:
Continuing from conversations in [the Alda Slack group](https://slack.alda.io), tracking the work to be done here:

* [x] Add ARM builds to `client/bin/build`
* [ ] Tune JVM options so that alda-player processes use less memory?
* [ ] Add arm64 release testing to CircleCI build (looks like they don't support 32-bit ARM, so can't test ARM 7)
* [ ] Update Alda API to include arm-7 and arm64 releases
* [ ] Ensure that `alda update` can find the right releases for ARM 7, arm64
* [ ] Ensure that ARM releases are included at alda.io/install

PR #521, opened 2026-05-08
Reduce default JVM memory limits for alda-player (fixes #410)

### LINKED ISSUE #410: Alda suddenly consume a huge amount of RAM
# 🐞 Bug report 🐞

## Description

When left open the alda's REPL for a very short amount of time, suddenly it consumes a huge amount of RAM at rapid speed (~1.5 gigabytes).

## Steps to Reproduce

1. Start `alda repl`
2. Leave open for a while

## Does the problem happen consistently?
Yes

## Is there anything else that it would be useful for us to know?
This also seems to happen when run `alda ps` and `alda-player info`. When I opened up `htop` it also shows that there are a lot of alda processes.

wth_alda

## Expected Behavior

The alda REPL starts normally and doesn't consume huge amount of RAM.

## Actual Behavior

The alda REPL starts normally, but then consumes a huge amount of RAM.

## Environment

OS: Ubuntu (Pop!_OS 21.10 x86_64)
Alda installed through Homebrew.

**Alda version:**

```
$ alda version
alda 2.2.0
```

```
$ alda-player info
alda-player 2.2.0
log path: /home/firefly/.cache/alda/logs
```

**Health check:**

```
$ alda doctor
OK  Parse source code
OK  Generate score model
OK  Ensure that there are no stale player processes
OK  Find an open port
OK  Send and receive OSC messages
OK  Locate alda-player executable on PATH
OK  Check alda-player version
OK  Spawn a player process
OK  Ping player process
OK  Play score
OK  Export score as MIDI
OK  Locate player logs
OK  Player logs show the ping was received
OK  Shut down player process
OK  Spawn a player on an unknown port
OK  Discover the player
OK  Ping the player
OK  Shut the player down
OK  Start a REPL server
nREPL server started on port 43725 on host localhost - nrepl://localhost:43725
OK  Find the REPL server
OK  Interact with the REPL server
OK  Shut down the REPL server
```

**Logs:**

```
```

SomedayToday:
## Summary

This PR decreases the default JVM heap memory settings for `alda-player` in order to prevent the excessive RAM consumption described in #410.

## Problem

As reported, when running Alda's REPL (or multiple player processes), `alda-player` can rapidly consume several gigabytes of memory. This is especially problematic on low-memory devices like the Raspberry Pi (#403).

The current default settings allow up to **1024 MB** of maximum heap (`-Xmx1024m`), which is unnecessarily high for a music player process.

## Changes

I have updated the `jvm_opts` variable in the player build script to use more conservative memory limits:

- `-Xms64m` (initial heap size – reduced from 256 MB)
- `-Xmx256m` (maximum heap size – reduced from 1024 MB)

The other JVM options (`-XX:+UseG1GC`, `-XX:MaxGCPauseMillis=100`, `-DlogPath=tmplog`) remain unchanged.

These values were chosen because:
- **64 MB** initial heap is enough for normal operation and keeps memory usage low at startup.
- **256 MB** maximum heap provides a safe ceiling while still allowing the player to handle complex scores.
- They prevent the player from claiming huge amounts of RAM even when many processes are running.

## Notes

- This change affects the build process (`bin/build-player` or similar script) and will apply to all future releases once merged.
- Users who need more memory (e.g. for very large orchestras) can still override the JVM options at runtime, for example:  
  `java -Xmx512m -jar alda-player-fat.jar …`
- The Windows build (via launch4j) automatically picks up the same JVM options, so no separate change is needed.

Closes #410

### DIFF
--- a/player/bin/build
+++ b/player/bin/build
@@ -49,7 +49,7 @@ non_windows_build="$build_dir/non-windows/alda-player"
 # Defining an initial value here means that if some logging sneaks in before we
 # can reconfigure the log path, a `tmplog` directory appears in the current
 # working directory, instead of a much uglier `${sys:logPath}` directory!
-jvm_opts="-XX:+UseG1GC -XX:MaxGCPauseMillis=100 -Xmx1024m -Xms256m -DlogPath=tmplog"
+jvm_opts="-XX:+UseG1GC -XX:MaxGCPauseMillis=100 -Xms64m -Xmx256m -DlogPath=tmplog"
 
 ### NON-WINDOWS BUILD ##############################
 


daveyarwood:
Thanks for the PR! This looks good to me. I'll test within the next week or so and confirm that performance is acceptable (at least on my machine 🙃) with the lower memory usage, and then we'll be good to merge.

daveyarwood:
Testing notes:

* I kept `htop` open to monitor my system memory usage. Baseline was ~4.2GB (out of 14.9GB).
* **Control study:** I ran `alda repl` with the current release version of alda (2.4.2). Memory usage increased to ~6.6GB. I ran `alda shutdown` and memory usage went back down to baseline. I ran `bin/smoke-test-example-scores` (with local modifications to use `alda play`), which plays each of the example scores in the repo one a time. While doing this, memory usage was ~7.5-8.0GB.
* I checked out this branch and ran `client/bin/build` and `player/bin/build` to build the processes, then ran `client/bin/run repl`. Memory usage was somewhat lower at rest, ~6 GB.
* I ran `bin/smoke-test-example-scores` the way it is in the repo (using `client/bin/run play`). While doing this, memory usage was ~6.5-7.0 GB.
* I didn't hear any issues with playback. No stuttering, glitching, etc.

The memory savings are modest (~1 GB), which doesn't really make a difference on my Thinkpad with 16 GB of RAM, but I can see how this would make a significant difference on memory-constrained devices. For example, in #410, the user's screenshot showed that they were using a device with only 4 GB of RAM, so 1 GB in memory savings would be meaningful there.

I think we're good to go on this. Thanks again for the PR!

PR #522, opened 2026-08-03
fix(repl): preserve trailing rests in REPL note timing via per-part s…

### LINKED ISSUE #421: Alda REPL "skips" rests at the end of a line

Samyra312007:
Fixes #421

Problem
In the REPL, a trailing rest at the end of an evaluated line of input was effectively stripped from playback. When the next line was submitted, its first note played too early for example, a quarter note early when the previous line ended in r4. Reproducible both via alda repl --client --port 43009 --message '{"op":"eval-and-play",...}' and interactively with Up+Enter.
Root cause
When transmitting an incremental REPL update, the client re-bases the new events' offsets so they follow the player's current position. That position was computed as the end of the last note sent to the player (transmitter.SyncOffset). Rests advance a part's CurrentOffset but produce no score event, so the rest at the end of a line was never included the player was told to start the next line at the end of the last note rather than the end of the last rest, collapsing the gap.
A naive fix advancing the sync offset by the trailing rest breaks multi-track scores where parts sit at different positions (e.g. piano: g a b > c | violin: c d e f), since a single global offset can't represent where each part's track left off.
Fix
Replaced SyncOffset with per-part sync offsets keyed by stable part ID (transmitter.SyncOffsets map[string]float64). For each part, the sync offset is the player's position on that part's track the end of the last transmitted event (notes and rests) for that part, clamped to the earliest offset of the incoming events for that part. This:
- Preserves gaps created by trailing rests (the reported bug).
- Keeps each part independent, so multi-part and voice-based scores sync correctly.
- Is keyed by part ID rather than pointer, since a part's pointer changes when a voice group ends but its ID is stable.
- Keeps the "play right away" behavior when new music starts at or before the player's current position (no regression of #415).
:play/:replay/:load/alda play are unaffected one-shot playback never passes sync offsets.
Tests
Added regression tests:
- client/repl/server_test.go — the exact #421 repro line asserts the second transmission starts 250 ms in, re-based to 250,750,…,3750; no-trailing-rest unchanged; rest-only line delays the next note; voices start immediately; player-behind-part clamps to 0; two-track divergence stays independent.
- client/transmitter/osc_test.go — trailing-rest rebase, negative-offset clamp, empty-map no-op.
- client/model/notes_test.go — documents the invariant that rests advance CurrentOffset without producing an event.
All client tests pass via go test ./...


daveyarwood:
Hi @Samyra312007, thanks for the PR! I have a few things on my plate at the moment, but I'm hoping to have time to review before the end of the week.

